PR #36, opened 2015-02-25
fixes to enum problems in TrackVehicle

jcmadsen:


rserban:
Justin, this still does not compile on Linux/GCC.  Have you tried it?

jcmadsen:
I don't have linux locally, so I'm relying on Travis CI to do that. 
How do I have Travis CI enable the TrackVehicle build, which also requries UNIT_IRRLICHT?

jcmadsen:
Radu, your fixes for linux work fine with VS 2012, so unless anyone finds anything else this is ready to go.

PR #38, opened 2015-03-20
Feature/gear geometry

jcmadsen:
First working version of custom collision detection between concave gear seat and cylinder shoe pin, new classes are in TrackCollisionCallback.h. 
Geometry for the gear and pin are defined analytically and parametrized in class GearPinGeometry.
The collision callback is implemented as a ChCustomComputeCollisionCallback in the newly created class GearPinCollisionCallback.
_contact_reporter was also created to output all contact data to console or a file, using a ChReportContactCallback implementation.
I have implemented an outward envelope for both the broadphase and narrowphase, but have not dealt with an inward safe margin yet.


jcmadsen:
rebasing into a few dozen essential commits.

PR #39, opened 2015-04-03
collision callback between gear/shoe pin works and is tested.

jcmadsen:
runtime and vehicle models for validation with M113 reports in doc from (Shabana, Wallin).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ addons:
   apt_packages:
     - libirrlicht1.7a
     - libirrlicht-dev
+    - libxxf86vm-dev
+    - freeglut3-dev
 compiler:
   - gcc
   - clang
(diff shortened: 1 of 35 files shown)


tasora:
Justin:
   I am adding a new feature to C::E , a custom collision algorithm for 
polyline vs polyline (where the polyline can have concave/convex fillets 
and segments, as in your gear. This will be meant as a '2d' collision 
between the two polylines, but I think it would suffice (as shoe pins & 
gear contact are always in a plane)

alessandro


Il 03/04/2015 16:28, Justin Madsen ha scritto:

jcmadsen:
Hi Alessandro
This would be good to test against the analytical geometry I'm using in my
collision callback. Performance and accuracy could be compared when using
more or less lines to represent the concave curve in gears and chain
systems.

Thanks
Justin

jcmadsen:
Alessandro,
Do you have or will you be making a demo that uses the new polyline
collision detection?

Thanks,
Justin

tasora:
Hi Justin,
   I'm planning to prepare a demo in 15 days.
Alessandro.


Il 07/04/2015 22:16, Justin Madsen ha scritto:
> Alessandro,
> Do you have or will you be making a demo that uses the new polyline
> collision detection?
>
> Thanks,
> Justin

PR #46, opened 2015-04-22
Validate ancf cable

jcmadsen:
New validation demo based on Alessandro's demo_FEAcables.
Validates results for 3D ANCF gradient deficient cable, as in  (Sugiyama, Mikkola, Shabana 2003),
added new constructor for ChBeamSectionCable.

### DIFF
--- a/src/demos/fea/CMakeLists.txt
+++ b/src/demos/fea/CMakeLists.txt
@@ -9,6 +9,7 @@ SET(FEA_DEMOS_IRR
 	demo_FEAthermal
 	demo_FEAelectrostatics
 	demo_FEAvisualize
+	validate_ANCFcable
 )
 
 #List all FEA demos that use both Irrlicht and Matlab
--- a/src/demos/fea/validate_ANCFcable.cpp
+++ b/src/demos/fea/validate_ANCFcable.cpp
@@ -62,11 +62,10 @@ int main(int argc, char* argv[])
 	const double EA = 1.26E4;	// [N]
   const double area = beam_diameter * beam_diameter / 4.0;  // cross section area [m2]
 	const double youngs_mod = EA / area;	// [N/m2]
-  const double area_I = EI / youngs_mod;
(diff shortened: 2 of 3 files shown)


rserban:
Justin,  it's not clear to me in what sense this is truly a validation program.  For that, it would need to compare the chrono results to something (analytical results, experimental data, simulation data obtained by some other means).

If you do have such data against which the chrono results can be compared, then the proper place for such a program is probably in the chrono-validation repository (which currently contains the validation of joints against ADAMS simulations, which used to be in chrono-T but were moved relatively recently).  In that case, please look at the utilities I put in the chrono-validation repository to support such file comparisons and see how you would use those for this code.  I can then look at how best to (re)organize the folder structure and cmake scripts there to accommodate such ANCF validation programs.

jcmadsen:
Strictly speaking there is no validation data to compare against at this point. Its a comparison against the model in the paper referenced in the demo, and the end goal is to output data to re-create the 4 plots referenced. This should go in the appropriate place where I can also upload the paper referenced, which is probably the new validation repo.

rserban:
We cannot validate against a paper and we wouldn't want to upload a paper in a github repository anyway.  Moreover, any validation test that we place in chrono-validation must return pass/fail, as these programs are run through CTest.

If we can get the data from the paper and use that to validate against, then this can become a proper validation program.  Otherwise it's just a demo that implements a model from some paper.

jcmadsen:
It's one of Hiro's papers, so I'm sure he won't be too upset about providing some of the time-domain data, specifically for Figs. 6, 7 8 and 9. In that case, it will be a matter of deciding of what is the exact definition of pass/fail in the case of time-domain data.

Further, I have my own set of Matlab code that implements this type of ANCF cable element, and could use it as a second form of validation for the demo.

rserban:
Let's then ask Hiroyuki for the data.  As for comparing time-domain data, there is support for that in the utilities of chrono-validation which I mentioned before (we obviously do this all the time when comparing against ADAMS results).

PR #47, opened 2015-04-23
removes bug that uses a deprecated function when calculating increments ...

jcmadsen:
...to the force and moment residual vectors.
Found when running demo_FEAvisualize.

### DIFF
--- a/src/unit_FEA/ChMesh.cpp
+++ b/src/unit_FEA/ChMesh.cpp
@@ -742,8 +742,12 @@ void ChMesh::VariablesFbLoadForces(double factor)
 		this->vnodes[in]->VariablesFbLoadForces(factor);
 
 	// internal forces
-	for (unsigned int ie = 0; ie < this->velements.size(); ie++)
-		this->velements[ie]->VariablesFbLoadInternalForces(factor);
+  ChVectorDynamic<> R_f(this->GetDOF());
+  this->IntLoadResidual_F(0, R_f, factor);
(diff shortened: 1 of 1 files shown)


andrewseidl:
Re-assigning @tasora, author and maintainer of `unit_FEA`.

tasora:
Justin: those  velements[ie]->VariablesFbLoadInternalForces(factor)  deprecated functions will be removed in future, please do not care about them. 

Btw the change that you did cannot work properly. Leave as it is.

Note: MOST of  the VariablesBlaBla... stuff in the entire code will be removed.
Currently, those VariablesBlaBla functions will be used only when using the INT_ANITESCU or INT_TASORA steppers (the default), but if you use all other new timesteppers, such as INT_HHT, they won't be used at all, so why care about them.

In most of my FEA tests, in fact, I am not using the old INT_ANITESCU. Just use the INT_EULER_IMPLICIT_LINEARIZED that is exactly the same old INT_ANITESCU, but it uses the new bookkeeping system. 

The reason why I haven't yet made INT_EULER_IMPLICIT_LINEARIZED as the default instead of the old INT_ANITESCU, is because there's still an issue with min-max limits. 
When fixed, this will be the default; then INT_ANITESCU and INT_TASORA will disappear, and also the old VariablesBlaBla stuff will be removed.

Alessandro

jcmadsen:
Sounds good, I will just change the integrator in the demos.

PR #48, opened 2015-04-23
Fix bugs in FEA demos from updating timesteppers.

jcmadsen:


jcmadsen:
would it be OK if I accepted this?

tasora:
yes, go ahead

jcmadsen:
hmm, I ran a re-base on the branch these changes were in, looks like that wasn't necessary

PR #49, opened 2015-04-27
streamOUT needs to be overloaded for ChSpmatrix type when writing to ChS...

jcmadsen:
...treamOutAscii, added this function to header and source

Add example of use to demo_LCPsolver.cpp
Max number of rows/cols printed hardcoded to 8


hmazhar:
why hardcoded to 8?

jcmadsen:
Same as whats in ChMatrix. But it should be part of the input parameters and just default it to 8. I will do that

PR #53, opened 2015-05-06
Ppt script demo m113

jcmadsen:
adding the output data and plotting capabilities needed to validate test_M113 against reported values in report by Shabana, Wallin.

### DIFF
--- a/src/demos/trackVehicle/ChronoTrack_pandas.py
+++ b/src/demos/trackVehicle/ChronoTrack_pandas.py
@@ -69,6 +69,40 @@ class ChronoTrack_pandas:
                     
         return out_minmax
     
+    # plot the chassis info
+    def plot_chassis(self, tmin = -1, tmax = -1):
+        # plot pos, vel, omega
+        # fig, axarr = plt.subplots(3,sharex=True)
+        fig, axarr = plt.subplots(2,sharex=True)
+        
+        # pull in data from files
+        inDat = ['time','x','y','z','Vx','Vy','Vz','Wx','Wy','Wz']
+        DF = pd.DataFrame(self._getDFhandle('chassis'), columns = inDat)
+        
+        # create the 3 subplots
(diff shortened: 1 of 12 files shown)


jcmadsen:
looks like Clang doesn't like C++11 lambda functions, so I'll remove that

andrewseidl:
> looks like Clang doesn't like C++11 lambda functions

Chrono itself does not currently support C++11, though I'm hoping we can change that for the next release.

I'm a bit surprised that GCC accepted the lambda fn...

jcmadsen:
Tossed the lambda function, combined that commit into the previous one.
What will surprise you most is that VS 2013 has decent C++ 11 support.

andrewseidl:
Please run [`clang-format`](http://clang.llvm.org/docs/ClangFormat.html#visual-studio-integration) on the modified C++ files - it looks like you're mixing 2-space and 4-space indents (Chrono uses the latter).

Is all the dead code necessary/can it be removed? For example, [trackVehicle/TrackVehicleM113.cpp#L612-L675](https://github.com/jcmadsen/chrono/blob/322a8104558c6cfaf233bfb6446fd0460361191c/src/demos/trackVehicle/subsys/trackVehicle/TrackVehicleM113.cpp#L612-L675)

@rserban: How does everything look to you?

rserban:
Looks fine to me.  As long as these pull requests only concern files in the demos/trackVehicle folder, the only concerns are to keep number of commits to the necessary minimum and to commit clean code.  All this functionality will eventually be superseded by the Chrono::Vehicle track support, at which point this entire folder in chrono/demos will be removed (hence the desire to keep the number of commits to a minimum).

jcmadsen:
that code is not dead, I just haven't gotten it working for an arbitrary number of runningGear systems.
And I though 2 space indents were being used, but I'll switch to 4.

PR #58, opened 2015-05-27
Added debug suffix in cmake configuration

ruisebastiao:
Generating different names for both build versions, avoids release/debug
mix problems

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -99,6 +99,8 @@ set(ENABLE_UNIT_TESTS    FALSE   CACHE BOOL   "Turn this ON to generate Chrono::
 # the compiler, generator, platform, etc....
 # Set a default build type if none was specified
 
+SET(CMAKE_DEBUG_POSTFIX "d" CACHE STRING "add a postfix, usually d on windows")
+
 if(MSVC OR XCODE_VERSION)
     set(CMAKE_CONFIGURATION_TYPES Release CACHE STRING "Choose the type of build.")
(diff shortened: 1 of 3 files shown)


andrewseidl:
Thanks for the contribution!

Could you describe the issue/use case that this solves? IDEs that support multiple configurations types (Visual Studio, Xcode) should be placing the libs/binaries in separate directories for each config. For generators that only support a single type, you usually maintain separate build directories (by hand).

-----------

On a different note, it appears that you have pushed some unrelated commits to the same branch as this PR. To fix this do:

    git checkout develop
    git branch chcascade-fix 5018f39
    git reset --hard HEAD~2
    git checkout chcascade-fix
    git cherry-pick -r bd08ec5 26bade4
    git push -u origin chcascade-fix
    git checkout develop
    git push -f

This will:

1. Create a new branch based off the commit prior to the debug suffix changes.
2. Roll back the last two (ChCascade-related) commits from the `develop` branch.
3. Move the ChCascade-relate commits to the new branch.
4. Push the new branch to GitHub.
5. Push the rolled-back `develop` branch to GitHub. The `-f` is required because you're essentially rewriting history.

ruisebastiao:
I am using Qt framework,  i was getting some wired problems during debug (i put a question ins stackoverflow http://stackoverflow.com/questions/30483089/problems-using-static-stdstring-in-shared-library) then i realize that both build configurations (Debug/Release)  where installed in the same target directory (target/lib). So using a suffix or create debug/build directories can avoid this situation in the future.

PR #62, opened 2015-07-29
Position solution vector option for HHT 

Bryan-Peterson:
This adds the position solution vector option for the HHT Integrator, adds a specific demo for ANCF, and includes ANCF Brick element.


rserban:
Bryan, please address the following issues and resubmit the pull request.  You may also want to look into *squashing* the commits in order to keep the log history clean (see `git rebase --interactive`).

1. The code does not compile on Linux.  This is due to the use of backslash as path separator (only valid under Windows); see line 20 of ChElementBrick.h.  
2. What is the reason for adding the member variable `Iter` to ChSystem (line 142 in ChSystem.h)?  The only reference to it is in a **demo** program (line 1375 of demo_FEAancf.cpp) and even there I don't understand its purpose.  Please revert ChSystem.h and modify demo_FEAancf.cpp accordingly.
3. The implementation of the 3 HHT options is **hardcoded** in ChTimestepper.  This needs to be changed to an option accessible through the public API.
4. Please remove comments in core classes and functions that reference ANCF tire models (or any particular application in general); see for example comment in ChTimestepper.h starting at line 454.
5. Is there any reason why ChElementBrick is entirely implemented in the header file?  This results in a *public* header file with 2200 lines!  Please move implementation to a cpp file.
6. This pull request includes an unnecessary modification to ChNodeFEAxyzD.h (just a commented code block).  Please revert this file.
7. Please spell check comments.

Thanks,
Radu

Bryan-Peterson:
Radu,


I will fix these error right away. I have a questions about item #5. What is the issue caused by implementing the element entirely in the header and will ChElementShellANCF need to be changed since it is also done entirely in the header?


-Bryan


________________________________

rserban:
At least a couple of reasons:

1. to separate interface from implementation -- a user of the library would like to look at the header file to see what functionality the library provides and not care much about how the library implements that functionality.
2. to optimize compilation -- with the implementation in the header file, compilation may take much longer; any small change somewhere deep in the implementation of a particular function will trigger recompilation of all units that include the header.

Bryan-Peterson:
Radu,

I've resolved the other issues with my implementation, but I'm not sure how to go about moving the element calculations to a .cpp file. Is it as simple as moving some of the larger functions into a .cpp file and including the header or is there more to it? I'm sorry for my difficulties with this, I'm a little unfamiliar with some of the fundamentals of C++.

-Bryan

________________________________

rserban:
There's a bit more to it, but that's the gist of it.
Looking at other classes in unit_FEA, I see that many of them are mostly (if not entirely) implemented in header files.  Since we'll want to revisit this for all of them eventually, just ignore this for now.

rserban:
Compilation on Linux still fails:

    /home/travis/build/projectchrono/chrono/src/demos/fea/demo_FEAancf.cpp: In function ‘void test_3()’:
    /home/travis/build/projectchrono/chrono/src/demos/fea/demo_FEAancf.cpp:793:10: error: ‘class chrono::fea::ChNodeFEAxyzD’ has no member named ‘SetFixedPos’

Is this code building fine on windows?

Bryan-Peterson:
Radu,

This is an oversight from tests I was running with a function I temporarily created in ChNodeFEAxyzD. It should just be "node ->SetFixed()". Sorry for the mix up. I will run the code right before submitting it next time.

-Bryan
________________________________

rserban:
That's good policy :-)
Make sure to compile in both release and debug modes.

Milad-Rakhsha:
1-The code does not compile on Linux. This is due to the use of backslash as path separator (only valid under Windows); see line 20 of ChElementBrick.h.

I had to change the line 20 from
#include "physics\ChContinuumMaterial.h"
to
#include "physics/ChContinuumMaterial.h"
and removed from /chrono/src/unit_FEA/ChElementBrick.h around line 587 the following line
assert (!section.IsNull());

to compile it on linux.

PR #70, opened 2015-09-30
Universal vehicle

jcmadsen:
pitman arm universal joint constrained axis is defined with hardpoints.

### DIFF
--- a/data/vehicle/generic/steering/PitmanArm.json
+++ b/data/vehicle/generic/steering/PitmanArm.json
@@ -28,9 +28,7 @@
 
   "Universal Joint":
   {
-    "Location":                 [ 0.129, 0.249, 0],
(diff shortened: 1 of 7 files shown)


rserban:
Justin, I am reverting this merge.  The changes to the Pitman arm template are unnecessary and result in a less general formulation.  Please submit another pull request with the changes to the suspension test rig. Thanks!

jcmadsen:
Radu,
In what way does inferring the universal joint constrained axis from the hardpoints in the template make the formulation less general? 
Thanks

rserban:
You have effectively hard-coded that the subsystem must be defined in a configuration in which the steering link is perpendicular to the plane of the universal joint's "cross".  That may be true for the one example we use right now, but may not be the case in general.

PR #78, opened 2015-12-16
Put the learning of the torque and rotation inside if statement so th…

armanpazouki:
…at the user can control it. Otherwise the previous default recording could cause memory overflow


armanpazouki:
the default behavior of ChLinkLinActuator is to record the torque and rotation of an actuator, in case learn is FALSE. For a small problem such as fork_lift, this does not cause major memory consumbtion; For one actuator, I assume you need to run the simulation for 1e6 steps to add 12MB to the memory (per actuator). Although this may look small for fork_lift, it can be significant when there is a lot of actuators in the system. I believe recording the actuator torque and rotation should be controllable by the user. The pull request lets the user to do so.

PR #83, opened 2016-05-01
Feature/wall clean

DavidHammen:
chrono does not compile -Wall clean. This branch addresses some of those warnings.


DavidHammen:
Ack. I didn't add Irrlicht and engine in my testing.

Working on a patch to my patch.

DavidHammen:
Finally!

My goal is to make chrono compile clean with compile flags `-Wall -Wextra -Wno-unknown-pragmas -Wno-unused-parameter` to gcc and clang. I'm not there yet, but I've reduced the result of `make |& wc -l` from almost one million lines to less than 36000 lines. The develop branch had a number of warnings, some of them very serious.

Metrics:

                       |   # lines    # warnings  # warnings    # types
                       |  from make    (total)     (unique)    of warnings
    -------------------+--------------------------------------------------
    develop            |   967267       87792        1352          30  
    feature/wall_clean |    35991        5583         654          20  


Key changes:

- I re-architected the serialization code somewhat as that's where some of the most egregious issues occurred. Clang was optimizing away some of the serialization code because of various undefined behaviors in that code. This change is not backward compatible. Implemented properly, it should have zero effect on the behavior of chrono.

- I added `override` wherever the compiler complained that a member was overriding but wasn't declared as doing so. The C++11 override specifier is a bit of a nuisance. Specify it for just one member of a class makes clang and gcc look to see if you should have used that specifier elsewhere in that class, and in classes that derive from that class. This change is invisible to the rest of chrono; it just removes some warnings. It has zero effect on the behavior of chrono.

- I hid some "-Woverloaded-virtual" by placing `#pragma GCC diagnostic push` and `#pragma GCC diagnostic ignored "-Woverloaded-virtual"` the declaration of functions that would otherwise receive this warning, and `#pragma GCC diagnostic pop` after. (There might be an MSVC equivalent, and that should probably be done as well.)  I only did this where it's obvious that the code intentionally hid the overloaded virtual function, such as the various `Copy` functions. I left a few of those overloaded virtual methods as-is as I'm not sure the overloading was intentional. Like the above, this change has zero effect on the compilation of or the behavior of chrono. It just removes some warnings.

- All told, I made quite a few changes:

        command_line_prompt> git diff  --shortstat develop
         330 files changed, 4970 insertions(+), 3985 deletions(-)

PR #90, opened 2016-07-13
Apply ChIrrApp::AssetBind/UpdateAll to visual assets which are held by physics items under assemblies

shigT:
(Sorry, I'm reposting because I mistook to branch.)

Current *ChIrrApp::AssetBind/UpdateAll* affect only physics items which are placed directly under *ChSystem*.
However, since programmers can create tree-like hierarchies with *ChAssembly* under the system, it would be natural that Binding/Updating affect whole these hierarchies.

This patch makes few changes in *ChIrrAssetConverter* to enable recursive search of visual assets which are held by physics items under assemblies.
A prevention of endless loops with circular references is included.

Thank you for considering.


rserban:
Hi Shig,
Thanks for the contribution!  I just merged the PR.
--Radu

PR #92, opened 2016-08-02
Contents of ChAssembly are not updated in certain situations such as ChSystem::DoStepDynamics()

shigT:
`ChPhysicsItem` has two overloaded update method: `Update(double mytime, bool update_assets)` and `Update(bool update_assets)`.
`ChAssembly` has an overriding of the second `Update()` to propagate method call to its contents, but not of first one.
Since some callers such as `ChSystem::DoStepDynamics()` take the first version of `Update()`, there are many chances to ignore updating assembly contents.
This causes some problems; for example, `ChLink` which contained in assemblies will not work.

This patch lets `ChAssembly` having an overriding of the first overloaded `Update()`, i.e. `Update(double mytime, bool update_assets)`, to fix above mentioned issues.

### DIFF
--- a/src/chrono/physics/ChAssembly.cpp
+++ b/src/chrono/physics/ChAssembly.cpp
@@ -616,6 +616,13 @@ void ChAssembly::Setup() {
         ncoords_w - ndoc_w;  // number of degrees of freedom (approximate - does not consider constr. redundancy, etc)
 }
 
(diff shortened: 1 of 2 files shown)


tasora:
Thank you for the pull request. Now merged. 
Note that this proliferation of different Update functions comes from old part of the code. My idea is that in future we should have just a single Update function signature, to avoid confusion.

PR #94, opened 2016-08-13
NarrowCallback should be able to overwrite the contact info

armanpazouki:
"const" does not allow it


rserban:
Arman, this callback is provided for reporting purposes; the user should not modify the contact info.  What do you have in mind?

armanpazouki:
Radu, thanks for looking at it. I did not know it is intended for reporting contact. At least it is a poor choice of the name, besides contacts can be reported in other ways as well.
What I would like to do is to rewrite some contact data for some primitives, rather than relying on bullet.

rserban:
Got it.  Of course, this would be best done before the Bullet narrow phase, but for now it's probably the easiest way to do that.

PR #96, opened 2016-08-27
Update usage of Submatrix and Subvector

TV4Fun:
Blaze has merged the {Sparse,Dense}Sub{matrix,vector} classes, so the build
is broken with new versions of Blaze. This fixes that by updating the usage
and also removes some files that referenced the Submatrix/vector classes but
weren't actually using them.

### DIFF
--- a/src/chrono_parallel/ChDataManager.h
+++ b/src/chrono_parallel/ChDataManager.h
@@ -40,11 +40,11 @@
 // Blaze Includes
 #include <blaze/math/CompressedMatrix.h>
 #include <blaze/math/DynamicVector.h>
-#include <blaze/math/DenseSubvector.h>
+#include <blaze/math/Subvector.h>
 
 using blaze::CompressedMatrix;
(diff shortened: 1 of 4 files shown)


rserban:
Joel, thank you for your contribution.

I merged the PR, but had to do it manually, as a few other changes were needed:

(1) with Blaze 3.0, clear() on a CompressedMatrix with zero capacity fails (I added the necessary checks)
(2) we are still supporting Visual Studio 2013 which, due to its limited C++11 support, requires Blaze 2.2 (I added the necessary checks for BLAZE_MAJOR_VERSION)

Cheers,
Radu

TV4Fun:
I'm aware of the issue with Blaze 3.0. This is fixed already in the development branch of Blaze (see https://bitbucket.org/blaze-lib/blaze/issues/47/segfault-on-clear-on-default-constructed) and will be fixed in Blaze 3.1.

PR #102, opened 2017-01-11
Add ChFunction support for generic lambdas

tjolsen:
Note: the new ChFunction_Lambda class requires -std=c++14 in order to be used properly since generic lambdas are not supported in C++11. This update was made explicitly in src/CMakeLists.txt in this commit, but should be achieved with conditional compilation.


tjolsen:
The unit test in this is bad. I had it returning 1 on success, not 0, so all of the builds are failing on TravisCI. Fixed and submitting new pull request.

PR #116, opened 2017-10-27
Feature/triad

malibrud:
This pull request adds a constructor to the ChCoordsys class.  The constructor generates the coordinate system based on a triad of points origin, xdir, ydir.  Origin becomes the position of the coordinate system.  The orientation is determined by the X axis being parallel to (xdir - origin).  the Z axis is parallel to X % (ydir - origin).  The Y axis is parallel to Z % X.  (Here % denotes the cross product).  

This constructor is useful in that it allows direct creation of a coordinate system uniquely determined by three points which can be generated from CAD or measured.  It is a natural way to setup a coordinate system.

I modified the ChCoordsys.h file and the utest_CH_coords.cpp file to test the code for typical and edge cases.  Some unit test cases were compared against independent calculation from either:

http://www.andre-gaschler.com/rotationconverter/

or Octave.

The method used when the trace(R) < 0 was leveraged from Martin Baker's site: http://www.euclideanspace.com/maths/geometry/rotations/conversions/matrixToQuaternion/

however I collapsed the last three cases based on a cyclic relationship between X, Y, and Z.

Mark Brudnak


rserban:
Hi Mark,

First of all, thank you for the contribution.  I agree that this would be a useful constructor for a ChCoordsys.

However, in order to eliminate duplicate code, it would be good to try to implement this using methods already available in ChMatrix33 (which is what we use to represent a rotation matrix).  In particular, Set_A_axis() can be used to set a rotation matrix from 3 mutually orthogonal unit vectors and Get_A_quaternion can be used to extract the corresponding quaternion (i.e. most of the work currently done in this new constructor).

Having said that, I suspect that such a simpler implementation may raise some issues with cyclic dependencies between template classes (ChCoordsys and ChMatrix).  I didn't get a chance to look more closely at this problem, also because I would like to eventually do some refactoring/redesign of several of the core low-level Chrono classes (including ChMatrix33) that would help with such issues.

I would also like to ask you to consider, for future contributions, the following: (1) please do not modify the top header comment in the Chrono files; and (2) please apply the format defined by the .clang-format in the root directory of the Chrono source.

Thank you. 
Best,
Radu

malibrud:
Radu,

Thank you for your suggestions.  I made the changes which greatly simplified the code.  The new constructor is as follows:

`explicit ChCoordsys(const ChVector<Real>& origin, const ChVector<Real>& xdir, const ChVector<Real>& ydir) {
		ChVector<Real> ux, uy, uz;

		// Get orthonormal vectors from points provided
		ux = xdir - origin;
		ux.Normalize();
		uz = ux % (ydir - origin);
		uz.Normalize();
		uy = uz % ux;
		uy.Normalize();

		ChMatrix33<> R;
		R.Set_A_axis(ux, uy, uz);
		rot = R.Get_A_quaternion();
		pos.Set(origin);
	}` 

Unfortunately, by including "chrono/core/ChMatrix33.h" a cascade of template errors are generated.  The first one generated is C2988 "unrecognizable template declaration/definition" and the offending code is as follows:

`    template <class RealB>
    ChVector<Real> Matr34_x_Quat(const ChQuaternion<RealB>& qua) {
        assert((rows == 3) && (columns == 4));
        return ChVector<Real>(Get34Element(0, 0) * (Real)qua.e0() + Get34Element(0, 1) * (Real)qua.e1() +
                                  Get34Element(0, 2) * (Real)qua.e2() + Get34Element(0, 3) * (Real)qua.e3(),
                              Get34Element(1, 0) * (Real)qua.e0() + Get34Element(1, 1) * (Real)qua.e1() +
                                  Get34Element(1, 2) * (Real)qua.e2() + Get34Element(1, 3) * (Real)qua.e3(),
                              Get34Element(2, 0) * (Real)qua.e0() + Get34Element(2, 1) * (Real)qua.e1() +
                                  Get34Element(2, 2) * (Real)qua.e2() + Get34Element(2, 3) * (Real)qua.e3());
    }`

I am not sure why including this file creates template errors. Perhaps this is the circular reference that you mentioned above.

Regarding the unit tests, I think that I would withdraw any changes to utest_CH_coords since the code is now very straight forward and I already tested the orthogonalization code.

Best regards,

Mark

rserban:
Hi Mark,

Yes, these compiler errors are likely due to cyclic dependencies between template classes (ChCoordsys and ChMatrix/ChMatrix33 in this case).  Unfortunately, I don't have time these days, but I will look into how this can be circumvented with the current implementations of these classses.

In the meantime, a couple of comments on your latest version of this constructor:
- there is no need for normalizing `uy` (as the cross product of two othrogonal unit vectors, it will already have unit norm)
- you should use `ChMatrix33<Real> R;`
- you could simply use `pos = origin;`
- I would declare and initialize the 3 vectors at the same time, using things like:
 `ChVector<Real> ux = xdir - origin;`

Best,
Radu

PR #123, opened 2018-03-30
Refactor out irrlicht dependency from chrono_vehicle

bholcomb:
Having Irrlicht dependent classes in the Vehicle library seems like an unnecessary dependency for users of chrono_vehicle, especially ones who want to use official binaries.   Pulling out the Irrlich classes into their own library breaks the dependency and allows me to build/run all the demos for lightweight testing, but still use the same vehicle library in my own system without the dependencies on irrlich (and it's dependencies, which I have different versions of).

This pull request puts all irrlicht specific vehicle classes into a new library called chrono_irrVehicle and updates all demos to use the new library.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -332,6 +332,7 @@ add_subdirectory(chrono_parallel)
 add_subdirectory(chrono_opengl)
 #add_subdirectory(chrono_ogre)
 add_subdirectory(chrono_vehicle)
+add_subdirectory(chrono_irrVehicle)
 add_subdirectory(chrono_fsi)
 
 # Add directories for model libraries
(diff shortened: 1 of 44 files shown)


rserban:
I understand your reasoning, but I'm not in favor of this solution (for several reasons).  This is a slippery slope.  Irrlicht is only one such possible dependency of Chrono::Vehicle (OpenCRG, HDF5, etc. raise the same issue and it's likely we will have more in the future).  
The solution here, especially for someone wanting to integrate Chrono into a 3rd party software, is to configure and build it from source, which allows you to not enable the Irrlicht module at all or else build it with your own dependencies.   
The pre-compiled binaries are provided as a convenience for users who want to quickly check out Chrono, but they are limited and limiting.

bholcomb:
Sounds like a plan.  thanks!

PR #125, opened 2018-06-25
fix for linux sockets, as suggested by @mmorandi

louisgag:
adding header file which was missing and correction for opengl

### DIFF
--- a/src/chrono_cosimulation/ChSocket.cpp
+++ b/src/chrono_cosimulation/ChSocket.cpp
@@ -282,7 +282,7 @@ void ChSocket::setSocketBlocking(int blockingToggle) {
 
 int ChSocket::getDebug() {
     int myOption;
-#ifdef TARGET_OS_MAC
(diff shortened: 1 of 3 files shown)


rserban:
Louis, I have a question about the modification you made to ChOpenGLBase.h (defining the macro GLM_ENABLE_EXPERIMENTAL).
Recently but before your PR was merged, Dario Mangoni modified the CMakeLists.txt used to build the Chrono::openGL library to define this through a compile definition (see commit f814e9a2).  
So now we get compilation warnings about macro redefinitions.
Did the mechanism provided by Dario not work for you?  Are you not using CMake to configure and build Chrono?

louisgag:
Hi Radu,
Apparently I had not pulled the latest version of CMakeLists.txt modified by Dario.
You can ignore my commit as it indeed creates a redundancy.
Sorry about this,
-Louis
PS: yes, I am using cmake and Dario's commit works fine

Issue #124, opened 2018-06-13
build failure with cuda 9.2


bartoszek:
Have some minor isse when building against new cuda (9.2) mainly in type conversion.
```
[ 96%] Built target demo_PAR_ballsSMC
Scanning dependencies of target demo_GL_cohesion
In file included from /opt/cuda/include/thrust/detail/reference.h:173,
                 from /opt/cuda/include/thrust/memory.h:25,
                 from /opt/cuda/include/thrust/device_ptr.h:25,
                 from /opt/cuda/include/thrust/device_malloc_allocator.h:25,
                 from /opt/cuda/include/thrust/device_vector.h:25,
                 from /home/bartus/AUR/chronoengine/src/chronoengine/src/chrono_fsi/ChFsiDataManager.cuh:24,
                 from /home/bartus/AUR/chronoengine/src/chronoengine/src/chrono_fsi/ChFsiInterface.h:22,
                 from /home/bartus/AUR/chronoengine/src/chronoengine/src/chrono_fsi/ChFsiInterface.cpp:18:
/opt/cuda/include/thrust/detail/reference.inl: In instantiation of ‘thrust::reference<Element, Pointer, Derived>::value_type thrust::reference<Element, Pointer, Derived>::strip_const_get_value(const System&) const [with System = thrust::cuda_cub::tag; Element = chrono::fsi::Real3; Pointer = thrust::device_ptr<chrono::fsi::Real3>; Derived = thrust::device_reference<chrono::fsi::Real3>; thrust::reference<Element, Pointer, Derived>::value_type = chrono::fsi::Real3]’:
/opt/cuda/include/thrust/detail/reference.inl:105:54:   required from ‘thrust::reference<Element, Pointer, Derived>::value_type thrust::reference<Element, Pointer, Derived>::convert_to_value_type(System*) const [with System = thrust::cuda_cub::tag; Element = chrono::fsi::Real3; Pointer = thrust::device_ptr<chrono::fsi::Real3>; Derived = thrust::device_reference<chrono::fsi::Real3>; thrust::reference<Element, Pointer, Derived>::value_type = chrono::fsi::Real3]’
[...]

PR #127, opened 2018-08-02
Fixed include paths

saullocastro:
While installing Chrono in my Ubuntu machine I realized that those fixes were needed.

### DIFF
--- a/src/chrono/collision/bullet/BulletCollision/BroadphaseCollision/btBroadphaseProxy.h
+++ b/src/chrono/collision/bullet/BulletCollision/BroadphaseCollision/btBroadphaseProxy.h
@@ -16,9 +16,9 @@ subject to the following restrictions:
 #ifndef BROADPHASE_PROXY_H
 #define BROADPHASE_PROXY_H
 
-#include "LinearMath/btScalar.h" //for SIMD_FORCE_INLINE
-#include "LinearMath/btVector3.h"
-#include "LinearMath/btAlignedAllocator.h"
+#include "../../LinearMath/btScalar.h" //for SIMD_FORCE_INLINE
+#include "../../LinearMath/btVector3.h"
+#include "../../LinearMath/btAlignedAllocator.h"
 
 
 /// btDispatcher uses these types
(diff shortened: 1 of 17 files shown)


rserban:
A quick question: how are you building Chrono?  If using the provided CMake-based system, this should not be an issue (see below).

The files you modified (bullet & HADC) were adopted in Chrono with some modifications.  However, the goal was to keep changes to these libraries at a minimum and only where absolutely needed for interfacing to Chrono.  As such, we didn't change the way header file includes were handled in the original sources. 

The solution to this was to ensure that the proper directories are added to those the compiler searches for include files.   With CMake, this is done with:
```
set(CH_INCLUDES "${CMAKE_CURRENT_SOURCE_DIR}")
set(CH_INCLUDES ${CH_INCLUDES} "${CMAKE_CURRENT_SOURCE_DIR}/collision/bullet" )
set(CH_INCLUDES ${CH_INCLUDES} "${CMAKE_CURRENT_SOURCE_DIR}/collision/gimpact" )
set(CH_INCLUDES ${CH_INCLUDES} "${CMAKE_CURRENT_SOURCE_DIR}/collision/convexdecomposition/HACD" )

include_directories(${CH_INCLUDES})
```
(see `src/chrono/CMakeLists.txt).

If you are not using the provide CMake system, you should be able to do the equivalent thing.

Having said that, the bullet and HADC files in Chrono already have modifications.  So if you provide us with a compelling reason, we will certainly consider adopting the changes you suggested.

saullocastro:
I built using the default cmake scripts that you provided. However, I did not use the official release 3.0.0, but the develop branch of this repository. After running cmake I ran make install. Once trying to build some of the examples I noticed this problem with the include files. Attached you can see the CMakeCache.txt file that I ended up using.

The only module I could not compile is "Chrono::FSI", giving the same error described [here](https://github.com/projectchrono/chrono/issues/124). 

Thanks for the attention!
[CMakeCache.txt](https://github.com/projectchrono/chrono/files/2253320/CMakeCache.txt)

saullocastro:
@rserban I closed after discovering that I was running CMake directly from the folder where the FSI examples are, thereby skipping the extra include directives you mentioned

PR #135, opened 2018-10-08
Diagnostics demo for chrono collision models

kishor8dm:
As requested, here is the demo for generating the diagnostics for the wrapped collision model.


tasora:
This demo may be removed now, because in my recent commit I implemented the functionality proposed by Kishor directly inside the ChronoEngine_Irrlicht module. In detail, there is a new function ChIrrTools::drawCollisionShapes(). 
Also, the GUI of ChIrrAppInterface has a new checkmark to enable the draw of collision shapes, just open th panel by pressing "i" and check the "Draw coll.shapes" box.

PR #146, opened 2019-03-28
Fix build with GCC on FreeBSD

pkubaj:
This is needed because some platforms still use GCC (like all powerpc platforms).

### DIFF
--- a/src/chrono_thirdparty/filesystem/path.h
+++ b/src/chrono_thirdparty/filesystem/path.h
@@ -35,7 +35,7 @@
 #endif
 
 #ifdef __FreeBSD__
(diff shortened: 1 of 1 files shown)


pkubaj:
Yes, this also works.

rserban:
Could you then amend your PR?  Or else I can push this fix.
Thanks

pkubaj:
Changed.

rserban:
I thought it's not worth pulling 2 commits for a single-line change, so I made and pushed this change directly.  Please let me know if this works for you.

pkubaj:
Yes, it works, thanks.

PR #159, opened 2019-05-20
mingw+windows patches

aluaces:
Please consider merging the following set of patches that fix some compilation errors that cannot be noticed when compiling for windows with VisualStudio.

I found them while cross-compiling from linux with mingw-gcc 8.3.


thepianoboy:
I've been meaning to dig into this for a while. I'm not sure we're concerned about cross-compiling from Linux to Windows, but we should probably at least make an attempt to get MinGW working as a native windows compiler.

I suggest that we merge this into a new feature branch and poke at it a little bit before merging into chrono/develop.

thepianoboy:
Buildbot is running the patch now. I'll keep an eye on the Windows builder results and poke at it once I'm back from vacation.

PR #160, opened 2019-05-22
WIP: Feature/chrono granular

conlain-k:
# Chrono::Granular Module
Adds Chrono::Granular GPU solver for systems of monodisperse spheres. This supports full frictional contact between spheres, analytical boundary conditions, and meshes through a force-displacement co-simulation.

Chrono::Granular requires CUDA and a GPU and uses Chrono's packaged version of CUB. This PR updates that packaged version of CUB as well.

This module is only supported on Linux (Arch and CentOS have been tested extensively). Windows is not supported due to poor CUDA support by Windows, especially for Unified Memory. The module will build on Windows but issues with the CUDA driver make it unable to run. If anyone manages to make Granular work on Windows, please send us a PR.

I recommend merging with commits squashed so that we can retain some of the history but avoid a bunch of unnecessary commits in the Chrono tree.

### Things to do before merge: 
- [x] check status of Windows Support
- [x] verify builds on Linux builders and compatibility with other modules
- [x] Make the mini test a true unit test and add to CTest
- [x] Add some metrics tests and projects to [chrono-projects](https://github.com/projectchrono/chrono-projects)
- [x] Make sure that the module is linkable by external projects


conlain-k:
Now has a true unit tests and two metrics tests provided by projectchrono/chrono-projects#1.

conlain-k:
Accompanying PR projectchrono/chrono-projects#1 has been merged. Currently, Chrono::Granular will build on most Linux platforms but **will not run on any graphics card Maxwell or older**. Cards Pascal or newer should work fine. A long term fix is to avoid building executables, or perhaps any chrono::granular binaries, on a system if a supported graphics card is not present. Until such a fix, all tests are pulled out of CTest to avoid breaking builds.

PR #197, opened 2019-11-07
Fix demo_WheeledGeneric when Irrlicht is not available

alhirzel:
There was a reference to Irrlicht that was not `#ifdef`'d. I also `#ifdef`d some of the visualization configuration to better show what is not needed when visualization is unused.

### DIFF
--- a/src/chrono_vehicle/terrain/GranularTerrain.h
+++ b/src/chrono_vehicle/terrain/GranularTerrain.h
@@ -186,7 +186,7 @@ class CH_VEHICLE_API GranularTerrain : public ChTerrain {
     unsigned int GetNumParticles() const { return m_num_particles; }
 
     /// Get the terrain height at the specified (x,y) location.
-    /// This function returns the heighest point over all granular particles.
+    /// This function returns the highest point over all granular particles.
     virtual double GetHeight(double x, double y) const override;
 
     /// Get the terrain normal at the specified (x,y) location.
--- a/src/demos/vehicle/demo_ArticulatedVehicle/CMakeLists.txt
+++ b/src/demos/vehicle/demo_ArticulatedVehicle/CMakeLists.txt
@@ -1,5 +1,6 @@
 #=============================================================================
 # CMake configuration file for the ARTICULATED_VEHICLE demo.
+# This example program requires Irrlicht run-time visualization.
 #=============================================================================
 
 IF(NOT ENABLE_MODULE_IRRLICHT)
--- a/src/demos/vehicle/demo_CRGTerrain/CMakeLists.txt
+++ b/src/demos/vehicle/demo_CRGTerrain/CMakeLists.txt
@@ -1,6 +1,6 @@
 #=============================================================================
 # CMake configuration file for the CRGTerrain demos.
-# These example programs requires Irrlicht run-time visualization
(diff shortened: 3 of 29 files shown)


rserban:
I'm not sure I understand the point of this change.   This demo is not built unless the Chrono::Irrlicht module is enabled (see its CMakeLists.txt).  The "generic" vehicle is used mostly as a sandbox for developing new vehicle features and general testing and so we do want this demo to always have visualization enabled.

alhirzel:
When I saw the issue that e4410d70d4c04bed517cb74e3d12aef20893149d fixes, I did a look-around to ensure it did not affect any other examples. In doing so, I noticed the following have the same spirit of being able to build without Irrlicht:

* `vehicle/demo_WheeledGeneric/demo_VEH_WheeledGeneric.cpp`
* `vehicle/demo_TrackedJSON/demo_VEH_TrackedJSON.cpp`
* `vehicle/demo_ISO2631/demo_VEH_Shock.cpp`
* `vehicle/demo_M113_Band/demo_VEH_M113_Band.cpp`
* `vehicle/demo_TrackedJSON_Band/demo_VEH_TrackedJSON_Band.cpp`
* `vehicle/demo_WheeledJSON/demo_VEH_WheeledJSON.cpp`

You raise an interesting point that these are not built when `ENABLE_MODULE_IRRLICHT` is unset; perhaps this behavior can be safely changed, so these demos are built whether or not Irrlicht is available. Would you mind if I addressed that in this PR as well? Alternatively, it seems like the build-time detection of Irrlicht is unmotivated, and perhaps it should be removed. (I would prefer to not do the latter, because I actually find this example very instructive.)

rserban:
You're right. There are some inconsistencies here. 
- some of these demos should not have the "no run-time visualization" option  (e.g., demo_VEH_WheeledGeneric and demo_VEH_TrackedJSON) because they are just a demonstration of the vehicle and really meant to be controlled through keyboard
- others (the "Band" tracked vehicles) have it because they are too computationally expensive to run close to real time and therefore interactive driving is not the best option
- others still have this option because they implement particular tests (e.g. demo_VEH_Shock), don't have an interactive driver, and thus could be run without visualization (for example in batch testing)

Having said that, it's great you want to help with a bit of house cleaning. 
- modify the corresponding CMakeLists.txt files to allow a particular demo to be run without Irrlicht if it already has the code for that (that is, eliminate the check for ENABLE_MODULE_IRRLICHT where appropriate).  Fix the demos as needed so that they built if Irrlicht is not enabled.
- but please **do not** change demos so that visualization assets are added conditional on Irrlicht being enabled!  This is because visualization assets may still be used for other rendering methods (such as postprocessing with POV-Ray or the likes).  Definition of visualization assets and actual run-time rendering are two different things.

alhirzel:
Good point on the visualization assets; that change is wrong, will revert. Thank you for the guidance!

alhirzel:
@rserban I'm happy with these changes, can you take another look? I tested by building with and without `MODULE_ENABLE_IRRLICHT`, and found+fixed a few issues.

rserban:
Looks good to me.  But please squash all these commits into a single one.

alhirzel:
squashed

PR #207, opened 2019-12-21
Provide more access to ANCF shell elements strain and stress vectors

recuero:
The purpose of this PR is to increase strain and stress information provided by the API for the ANCF shell elements. Example of how to use this method is shown in the ANCF shell 8 demo. The user can now obtain any stress or strain component by prescribing the integration point and the layer number.


recuero:
@rserban Thanks for the suggestions. They should be addressed in the updated PR.

PR #208, opened 2019-12-27
Minor performance and functionality updates

keonjoe:
Improved performance of updateFrictionData by ~2x (profiled with nSight compute). Updated .gitignore to exclude .vs/ directories generated by MSVS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ develop-eggs
 lib
 lib64
 __pycache__
+.vs
 
 ## Installer logs
 pip-log.txt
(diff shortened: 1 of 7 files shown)


keonjoe:
Since the original pull request, I have also added and tested a functionality to search bodies in the system by their ID.

rserban:
Could you please separate these two different things and have a PR for each?
(that's a good rule of thumb for PRs in general)
I could pull in your changes to add searching bodies by identifier.  However, the changes to Chrono::Granular will have to be reviewed by the main developers of that module.

PR #295, opened 2021-02-12
Fix MUMPS libraries discovery: find_library doesn't find libraries when "lib" prefixes are included in names

yurivict:


dariomangoni:
@yurivict with which Makefile/CMakeLists are you building Mumps?
Are you, by any chance, using [scivision/mumps](https://github.com/scivision/mumps)? If so, I would be really happy BTW.
Anyway, in the _official_ Mumps Makefile the naming is with the `lib` prefix, e.g.
```
mumps_lib:    $(incdir)/mumps_int_def.h \
              $(libdir)/libmumps_common$(PLAT)$(LIBEXT) \
              $(libdir)/lib$(ARITH)mumps$(PLAT)$(LIBEXT)
```
so I would suggest you to not delete the `lib`-prefixed libraries names, but to use the `find_library` with the `NAMES` options instead, so to allow to look for _both_ naming standards, e.g.
```
find_library(MUMPS_ONLY_LIBRARY_COMMON
			 NAMES "libmumps_common" "mumps_common"
			 PATHS ${MUMPS_ROOT}
			 PATH_SUFFIXES "lib" "lib64" "libraries"
         )
```
Because of this _I wouldn't_ merge this PR as is.

A separate comment applies to the `PATH_SUFFIXES` directory that you changed for the Mumps _shared_ library version.
Are you aware of any Makefile/CMakeLists that puts _shared_ libraries into `lib`?

I think that GNU and FHS standards tell to put any library (included shared libraries) in `lib`, but there are different way of thinking; Chrono (as many others), for examples, installs shared libraries into `bin` and import libraries (let's call them static) into `lib`. This is also suggested by [CMake's `INSTALL` command](https://cmake.org/cmake/help/latest/command/install.html?highlight=install) that clearly tells that shared libraries go in the runtime directory i.e. `bin`. That's why our Mumps CMakeLists does so.
BTW, even by using _scivision/mumps_ the shared libraries will fall into _bin_.

So, if you know of any Mumps Makefile/CMakeLists that produces shared libs into `lib` than I would rather _add_ `lib` to `PATH_SUFFIXES` instead of _replacing_ `bin` with `lib`.

It would be great if you have time to fix this few things, but obviously we can do it ourselves too!

yurivict:
I install MIMPS using the FreeBSD port ```math/mumps``` that uses the upstream URL http://www.code-aster.org/FICHIERS/aster-xxx.tar.gz

I experimented and found that cmake's ```find_library``` doesn't find shared libraries when the name has the ```lib``` prefix included in it.

PR #318, opened 2021-06-08
Fixed IISPH related issue:

Milad-Rakhsha:
- Fixed fsi demos to work properly with the IISPH method
- Fixed a bug in the IISPH implementation
- Fixed some compilation warnings


Milad-Rakhsha:
In general, a thorough cleaning should be performed on chrono fsi source codes to ensure that research-related and dead codes are removed from the develop branch

PR #320, opened 2021-06-10
Timoshenko Develop V1.0

pchaoWT:
1. A corotated Timoshenko beam is developed, and could be used for wind turbine blade modelling successfully.
2. The Timoshenko beam for fully material stiffness matrix (TimoshenkoFPM) cannot run correctly now, because the classical 3-dimensional shape function matrix of Timoshenko beam is not campitable with FPM.
3. A little change has been made for IGA beam to allow to set FPM.


pchaoWT:
The codes are not complete.

PR #338, opened 2021-10-03
Fix demo asset and output locations

lucasw:
Fixes a `Marder` vs. `marder` capitalization issue, rock.jpg is in a subdirectory, and demo output intermediate directories need to be created before lower level dirs.

Instead of renaming the Marder dir, maybe the preference would be for changing every place in code that assumes it is lowercase?

### DIFF
--- a/src/demos/irrlicht/demo_IRR_cohesion.cpp
+++ b/src/demos/irrlicht/demo_IRR_cohesion.cpp
@@ -142,7 +142,7 @@ void create_some_falling_items(ChSystemNSC& mphysicalSystem) {
 
         // optional, attach a texture for better visualization
         auto mtexture = chrono_types::make_shared<ChTexture>();
-        mtexture->SetTextureFilename(GetChronoDataFile("rock.jpg"));
+        mtexture->SetTextureFilename(GetChronoDataFile("textures/rock.jpg"));
         mrigidBody->AddAsset(mtexture);
     }
 
--- a/src/demos/irrlicht/demo_IRR_soilbin.cpp
+++ b/src/demos/irrlicht/demo_IRR_soilbin.cpp
@@ -123,7 +123,7 @@ class ParticleGenerator {
             auto cubeMap = chrono_types::make_shared<ChTexture>();
             cubeMap->SetTextureFilename(GetChronoDataFile("textures/concrete.jpg"));
             auto rockMap = chrono_types::make_shared<ChTexture>();
-            rockMap->SetTextureFilename(GetChronoDataFile("rock.jpg"));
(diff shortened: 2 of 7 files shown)


rserban:
I am a bit confused.  The change in capitalization for the data subdirectory now leads to demo_VEH_Marder not finding the assets (under Linux).: the Marder vehicle model points to OBJ files in the subdirectory "Marder", not "marder".
I don't have any issues running demo_VEH_Marder with the current code in the 'develop' branch, on either Windows or Linux.  What kind of issues were you trying to address with this name change?

Having said that, some of the fixes in your PR are indeed needed (e.g., the creation of the top-level output directory).  By the way, while it would be nice for create_directory to automatically create all necessary directories, I did not change the 3rd party 'filesystem' library we are using for that, because I wanted to keep it as close as possible with the std::filesystem library available in C++17 (so that we can seamlessly switch to using that if/when we move to a newer C++ version).

lucasw:
> I am a bit confused. The change in capitalization for the data subdirectory now leads to demo_VEH_Marder not finding the assets (under Linux).: the Marder vehicle model points to OBJ files in the subdirectory "Marder", not "marder".
I don't have any issues running demo_VEH_Marder with the current code in the 'develop' branch, on either Windows or Linux. 

Yeah I did that wrong, didn't see the obj paths.  The issue is that the controller json file paths for demo_VEH_MarderObstacle and a few others are in lower case:

https://github.com/projectchrono/chrono/blob/develop/src/demos/vehicle/tracked_models/demo_VEH_MarderObstacle.cpp#L82-L83

https://github.com/projectchrono/chrono/blob/develop/data/vehicle/Marder/driver/SpeedController.json

The obj files are consistent, are in the capitalized location:

https://github.com/projectchrono/chrono/blob/develop/src/chrono_models/vehicle/marder/Marder_Idler.cpp#L50-L51

It's a more minimal change to uppercase Marder in those json paths so I've updated the PR here.

rserban:
Lucas, I merged your PR.  Thank you for the contribution!

I'm curious to know if you're interested in particular in the Marder vehicle.  The reason I ask is because I intend to move several of the Marder demos out of the main Chrono repository and only keep demo_VEH_Marder.  Indeed, the other tests using the Marder vehicle are more involved and complex than expected from a simple illustrative demo.  Based on our current needs and work involving this model, I will move all other "demos" in a fork of the Chrono repository.   Please let me know if you are doing any work with any of these codes and I will let you know where they will be available.

lucasw:
> I'm curious to know if you're interested in particular in the Marder vehicle.

No I was just trying out all the demos to see what the library is capable of.  I am interested in construction and agricultural vehicles- but it looks like I can adapt the vehicle bases here easily enough to make my own crude versions.

PR #402, opened 2022-06-19
Fixed missing calls to SetVisualMaterial() causing textures not to sh…

### LINKED ISSUE #401: Textures are not showing up for .OBJ files that contain them.
As can be seen in for example the City Bus Demo that is part of the latest 7.0.3 release, but also in many other demos, loading an .OBJ file does not render any of its textures. This seems to happen for all objects that are loaded as ChTriangleMeshConnected.

image

A workaround (as suggested by @rserban on the mailing list) is to let irrlicht load them by loading them as ChObjFileShape (which apparently basically lets irrlicht load them instead of chrono). For reference, this bus should show up somewhat like this:

image

m4rr5:
…ow up. This fixes #401 and was tested by running one of the demos and both testing a custom .OBJ file for a vehicle and for the ground mesh.

### DIFF
--- a/src/chrono_irrlicht/ChVisualSystemIrrlicht.cpp
+++ b/src/chrono_irrlicht/ChVisualSystemIrrlicht.cpp
@@ -735,6 +735,7 @@ void ChVisualSystemIrrlicht::PopulateIrrNode(irr::scene::ISceneNode* node,
             mproxynode->Update();  // force syncing of triangle positions & face indexes
             mproxynode->drop();
 
+            SetVisualMaterial(mchildnode, shape);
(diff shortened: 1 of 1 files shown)


rserban:
Marcel, thanks for looking into this.
  
I merged your PR and, while it solves some cases, I don't think it's the full solution.  Indeed, for objects with multiple meshes and materials not all of them show up.  

To see that, modify the Gator vehicle's chassis (set m_has_mesh to true, instead of m_has_obj) and then run demo_VEH_Gator.

The reason I did not set the Irrlicht material on the mesh node (like you just did) was because I am already setting Irrlicht materials for each buffer of the mesh (see line 265 of ChIrrNodeShape.cpp).  Obviously that's not enough and your change may help point to the solution.

PR #406, opened 2022-06-28
Disable realtime step timer in vehicle by default.

m4rr5:
As mentioned on the forum, I think it's a bug to have a ChRealtimeStepTimer enabled by default that will wait on every invocation of Advance() to run in soft real-time. This is done too in the main() loops of all included demos (which is where it does belong). This PR disables the timer in ChVehicle. Arguably it could also be removed completely.


rserban:
Marcel, I am unable to merge this PR. I believe this is because you continued working in a branch from which I already merged a previous PR.

Anyway, while I disagree that this is a "bug", I concede that it may be better to initialize this flag to false.  
I will push this change myself, together with changing demos to use this ChVehicle option instead of doing their own soft real-time enforcing.

PR #407, opened 2022-06-30
Initialized real-time force to false in constructor.

m4rr5:
This should be a clean PR, replacing the one I did before. It will disable the code that will force the vehicle to run in real-time, leaving that to whatever "main" loop that it's being called from.


m4rr5:
Closing this as @rserban already fixed this in develop.

PR #411, opened 2022-07-09
Add an extra check to make sure we don't reference a non-existent member.

m4rr5:
The code that tries to remove visualization assets is apparently sometimes invoked before there are any, leading to the loop that was patched referencing members that don't exist (index out of bound). Added an extra check to the for loop to ensure that can no longer happen.


rserban:
Marcel, thanks for spotting this.
Could you please amend the fix to just:
```
 for (auto& cyl : m_cyl_shapes)
        ChPart::RemoveVisualizationAsset(m_wheel->GetSpindle(), cyl);
```
There's no need to check `id < GetNumDiscs()`

PR #435, opened 2023-01-18
Fix deletion of uninitialized pointer

jrbaxter0:
If a `ChVisualSystemIrrlicht` is constructed, then destructed without initializing, a segfault is caused due to `delete`ing the uninitialized `m_receiver` pointer. This fixes the problem.

It might be more elegant to instead initialize `m_pointer` with `nullptr` then check that here, but this seemed less invasive.


rserban:
Thanks for the fix.! I actually would prefer initializing m_receiver to nullptr.  Would you mind amending the PR?
Oh, and while you're at it, please move the definition of the destructor right after the constructor. :-)

jrbaxter0:
Should be all set

PR #450, opened 2023-04-26
MF Tire 5.2/6.1/6.2

MarvinSt:
This is a loose port of the [MFEval](https://www.mathworks.com/matlabcentral/fileexchange/63618-mfeval) open source library for Matlab/Simulink to C++ and integrated with Project Chrono. I am not a very experienced C++ developer, so I can't guarantee good code quality. I tried to keep the style as close as possible to the prior Pacejka model implementations of Chrono.
I validated the tire characteristics and they should produce identical results to the MFEval lib. I also tried to compare all equations against various TNO and ADAMS documents, but it is not always obvious which equations are actually used, described and implemented.
There is also a JSON parameter file loader included and I converted the sample files from MFEval to this JSON format.
I also updated one of the existing examples (vehicle wheeled json) to demonstrate the solution.

There are some issue / things to review / improve:

- [ ] Relaxation behaviour and low speed behaviour is not correct I believe, it needs to be reviewed
- [ ] Reverse driving is supported, but I flip certain signs and I am not sure if it is all fully correct
- [ ] Some calculations can be improved by collapsing the different functions (calc fx, fy, mz, etc.) into a single function, as separating these out will yield some redundancy and requires intermediate variables to carry over pre-calculations from other functions
- [ ] Turn slip is not actually implemented, although the foundation to support this is in place. It should be quite trivial to add.


rainergericke:
Thanks for the pull request. I will review it as soon as possible. Beginning on Friday this week.

rainergericke (approved):

rainergericke (approved):

PR #456, opened 2023-07-14
removing unneccesary if else statement

WangHanSolo:
**Motivation:** Cleaning up redundant code in the CMakeLists.txt in root directory.
**What was done:** Removed if else statement that seems extra.
**How it was tested:** Did not test as there is no functional change afaik.
**How to review:** Please let me know if I am seeing things incorrectly

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,11 +179,7 @@ endif()
 
 #-----------------------------------------------------------------------------
 
-if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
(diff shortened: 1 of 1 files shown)


rserban:
Please do not conflate PRs.  

Regarding the change to CMakeLists.txt:
- you are replacing an if-elseif construct and not an if-else, so you must check that this does not have unintended consequences on other platforms (e.g., Windows)
- we should consider removing CPack support altogether, which would make your change moot

WangHanSolo:
Closed after discussing with Radu offline.

Issue #270, opened 2020-10-10
Create Conan package [Feature Request]


jellespijker:
**Feature Request**
I think you guys are doing a great job and I really like to use Project Chrono in a variety of projects, But not every project of mine requires the same options. Compiling and maintaining different builds of dependencies for different projects can be difficult and time consuming and lets be honest who likes building and compiling dependencies.

I came across Conan (https://conan.io/) a year ago, which is a decentralized and multi-platform package manager, which allows me and consumers of my projects, to manage and share native binaries. This allows me to import dependencies such as Project Chrono for a new project with just a few lines in my CMake file or with a Python script. The beauty is that you have only to compile it once for a certain set of options and that these binaries can be uploaded to a server (public or private). If I have a different project with the same options, the binaries can be retrieved, saving a lot of time.
The same goes for all dependencies of Project Chrono, these are download or compiled as needed.

I have create my own repository which allows me use Project Chrono with Conan ( https://github.com/jellespijker/conan-projectchrono ), but this project is only tested on one platform (Linux - arch-based) and not every option and optional dependencies are included. I think it would be great if you guys provide support for a Conan package. This would i.m.o. greatly increase the ease with which I and others can use Project Chrono in our projects.

**Implementation as wrapper**

The conanfile.py from wrapper currently has the following content, and can used by other project by adding the FFF remote https://bintray.com/jellespijker/FantasticFuckinFusion

```python
from conans import ConanFile, CMake, tools
import os

class ProjectChronoConan(ConanFile):
    name = "projectchrono"
    version = "5.0.1"
    description = "C++ library for multi-physics simulation"
    topics = ("chronoengine", "project chrono", "multi-physics simulation", "DEM", "FSI", "vehicle", "simulation")
    url = "https://github.com/jellespijker/conan-projectchrono"
    homepage = "https://github.com/projectchrono/chrono"
    license = "AGPL3"
    # exports_sources = ["CMakeLists.txt", "patches/*.patch"]
    generators = "cmake", "cmake_find_package", "cmake_paths"

    settings = "os", "compiler", "build_type", "arch"
    options = {"shared": [True, False],
               "fPIC": [True, False],
               "openMP": [True, False],
               "TBB": [True, False],
               "HDF5": [True, False],
               "benchmarking": [True, False],
               "demos": [True, False],
               "tests": [True, False],
               "unit_cascade": [True, False],  # Not yet implemented
               "module_cosimulation": [True, False],
               "module_distributed": [True, False],
[...]

PR #459, opened 2023-07-16
Fix redefinition warning when using omp without CUDA

jellespijker:
Compiling with Thrust and omp without CUDA results in redefinition warning of `THRUST_DEVICE_SYSTEM` this PR add an extra check such that it is defined once


jellespijker:
Came across this while working on https://github.com/projectchrono/chrono/issues/270

Ported this PR as a patch for the CCI Conan Recipe: https://github.com/Ultimaker/conan-center-index/commit/32d988cc3db3d40cd959bfb17b837c3351a660a4

PR #477, opened 2024-01-27
Replace sprintf with snprintf (clean up compiler warnings)

jtkmckenna:
Clean up compiler warnings by replace sprintf with the safer snprintf

Build tested with clang 15 on MacOS 14.3

Since around clang 15, The warning: `'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]` fills the build log for this project. Hiding more meaningful warnings. Instead of supressing this warning, I propose we replace sprintf with snprintf.

Running synthetic tests using quick bench (clang 17, -O3): https://quick-bench.com/q/KtpfzBOCLkLi3vITdIB-gVxSVy4
```
const int BUF_SIZE=20;

static void sprintf_local_buf(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  int i = 0;
  for (auto _ : state) {
    char buf[BUF_SIZE];
    sprintf(buf,"%lu",(unsigned long)i);
    i++;
  }
}
// Register the function as a benchmark
BENCHMARK(sprintf_local_buf);

static void snprintf_local_buf(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  int i = 0;
  for (auto _ : state) {
    char buf[BUF_SIZE];
    snprintf(buf, sizeof(buf), "%lu",(unsigned long)i);
    i++;
  }
}
// Register the function as a benchmark
BENCHMARK(snprintf_local_buf);
```
Performance is unchanged:
image

===================================================
Possibly future solutions:

Exploring further, in many of the cases where we just put an integer into a buffer, would we prefer to use std::to_string()? In my synthetic tests it seems faster that snprintf? 

https://quick-bench.com/q/9oQLoXWX45qsF1jgXcrgY5Qatoc

Not entirely sure why, even when one pre-allocates the buffer std::to_string is faster, perhaps its range checking the input integer and can create a smaller buffer? Or the compiler has optimized out our actual test, would be worth benchmarking the impact in the project

```

const int BUF_SIZE=20;

static void sprintf_local_buf(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  int i = 0;
  for (auto _ : state) {
    char buf[BUF_SIZE];
    sprintf(buf,"%lu",(unsigned long)i);
    i++;
  }
}
// Register the function as a benchmark
BENCHMARK(sprintf_local_buf);

static void sprintf_preallocated_buf(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  int i = 0;
  char buf[BUF_SIZE];
  for (auto _ : state) {
    sprintf(buf,"%lu",(unsigned long)i);
    i++;
  }
}
// Register the function as a benchmark
BENCHMARK(sprintf_preallocated_buf);

char global_buf[BUF_SIZE];
static void sprintf_global_buf(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  int i = 0;
  for (auto _ : state) {
    sprintf(global_buf,"%lu",(unsigned long)i);
    i++;
  }
}
// Register the function as a benchmark
BENCHMARK(sprintf_global_buf);

static void snprintf_local_buf(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  int i = 0;
  for (auto _ : state) {
    char buf[BUF_SIZE];
    snprintf(buf, sizeof(buf), "%lu",(unsigned long)i);
    i++;
  }
}
// Register the function as a benchmark
BENCHMARK(snprintf_local_buf);


static void _to_string(benchmark::State& state) {
  // Code before the loop is not measured
  int i = 0;
  for (auto _ : state) {
    std::string buf = std::to_string((unsigned long)i);
    i++;
    // Make sure the variable is not optimized away by compiler
    benchmark::DoNotOptimize(buf);
  }
}
BENCHMARK(_to_string);
```
image


rserban:
I agree that this is a good change and I appreciate the initiative!

Having said that, many of these issues in Chrono are due to legacy code. We strive to periodically go back and update the code base to use more modern C++, but that's a long, ongoing process.  

In this particular case, I believe the solution is multi-faceted:
- in some cases, the API must be changed so we completely eliminate the need for `sprintf`/`snprintf`.
- in other cases, using std::to_string is the appropriate approach (you will see that we do use that in newer code). However, to_string should probably be limited to integral types since it may produce weird results for floating point types.
- in other places, it's simply more convenient to use `sprintf` (e.g., where formatted output is desired) and those places should be converted to use `snprintf` .

Here's what I suggest:
- give me a day or two to go through the code and make some changes that would completely eliminate the use of `sprintf`.
- at that time, you can update your PR to convert the remaining uses of `sprintf` to `snprintf`.
- note that there are more uses of `sprintf` than those your current PR fixed.  It'd be great if you could enable more of the Chrono modules and make these replacement throughout (some of the demos also use `sprintf`).  
- please do not make these replacements in any of the code under:
   -  *src/chrono_thirdparty* 
   -  *src/chrono/collision/bullet*

How does this sound?

rserban:
@jtkmckenna - I pushed the changes I mentioned above.  
It would be great if you could update your PR to replace the remaining instances of `sprintf` with `snprintf` and resubmit.
Thanks!

jtkmckenna:
This is a great proposal. I'll resolve the conflicts then proceed as planned

rserban:
Thanks @jtkmckenna !

One more request: we use a particular formatting (specified through the `.clang-format` file in the top-directory of the Chrono source). I don't know what IDE you use, but I'm sure there's a way to apply that formatting to the files you modified.

Could you please do that and add another commit to this PR?  I'll merge it after that.

--Radu

jtkmckenna:
Great point, auto-formatters are wonderful, I'll give it a pass over the files I've touched

Shall I get the very last instances of `sprintf` by grepping around the repo? They will be beyond what I can immediately test as I've not got all the project dependencies to build every last bit (Ok, I've started this, should it be its own PR since I've not tested these last ones)

rserban:
Maybe best for one of us to make the corresponding changes in modules that you cannot test.

jtkmckenna:
I've formatted all the files I've touched in this PR... 

(I used a diff to the main to check which files I needed to format... mostly making notes for myself here:
```
for i in `git diff --name-only main`; do
    clang-format -i -style=file ${i}
done
```

jtkmckenna:
> Maybe best for one of us to make the corresponding changes in modules that you cannot test.

Very good, I left it in a branch to sleep in my own fork: https://github.com/jtkmckenna/chrono/tree/sprintf_to_snprintf_unreached

PR #483, opened 2024-03-05
Optimized ChISO2631_Vibration_SeatCushionLogger absorbed power calculation

JKaniarz:
The GetAbsorbedPowerVertical() function is using 15% of total compute (on my particular run).  This is for two reasons:

1. It's building a std::vector of intermediate results even though it only uses the final result. (easy fix)
2. It retreads the entire history every time making the compute requirements O(n^2) with time.

In this patch I change the filters so they only update once as the data comes in. 

This change will result in different results due to what I assume is a bug in the current version. The code doesn't reset m_filter_abspow each pass through the array so some residual from the previous calculation affects the current result. This patch also fixes this.


rserban:
John, thanks for the contribution

PR #501, opened 2024-06-12
Remove extraneous register() from chrono_import

knauth:
This extra function call causes an error stating that the add-on is already registered when adding it via the GUI menu. Removing it resolves the issue.

We are also working on Blender 4.1 compatibility for this add-on; auto smoothing was removed and will need to be replaced with other smoothing functions.


tasora:
The register() call slipped in while doing debugging of the add-on. Removed.

PR #524, opened 2024-10-09
gitignore for visual studio 2022

ian-sixwheel:
adds gitignore entries for the build directory and various project files auto-generated by Visual Studio 2022 when following the build tutorial here: https://api.projectchrono.org/development/tutorial_install_chrono.html


rserban:
You should not do an in-source build to begin with.  So, no need for these additions to `.gitignore`.
The build tutorial you link to does not have the build directory (`chrono_build` in there) as a subdirectory of the source directory. Furthermore, the name `chrono_build` is just a suggestion; users are free to use whatever build directory name they want.

PR #525, opened 2024-10-09
Kraz demo realism fixes (fifth wheel and COM changes)

ian-sixwheel:
In the Kraz Demo, the Hitch being a Spherical joint is incorrect because a tractor trailer uses a "fifth wheel" hitch that only has 2 rotary degrees of freedom. The hitch can rotate in the Z direction, and can rock in the y direction, but cannot rotate in the x direction: 
image

As such, the spherical joint is not appropriate. This is however consistent with a Universal joint rotated 90 degrees such that the z axis of the joint is coincident with the x axis of the tractor. This PR creates a subclass of `ChChassisConnectorHitch` called `ChChassisConnectorFifthWheel` and replaces the standard hitch (spherical joint) with a Fifth Wheel hitch (universal joint).

In addition, the centers of mass of both the tractor and the trailer are unrealistically low, so I changed these to be more consistent with where they are in reality

### DIFF
--- a/src/chrono_models/vehicle/kraz/Kraz_tractor_Chassis.cpp
+++ b/src/chrono_models/vehicle/kraz/Kraz_tractor_Chassis.cpp
@@ -30,7 +30,7 @@ namespace kraz {
 const double Kraz_tractor_Chassis::m_body_mass = 10000.0;
 const ChVector3d Kraz_tractor_Chassis::m_body_inertiaXX(3441, 28485, 29395);
 const ChVector3d Kraz_tractor_Chassis::m_body_inertiaXY(0, 0, 0);
-const ChVector3d Kraz_tractor_Chassis::m_body_COM_loc(-2.0, 0, 0.6);
+const ChVector3d Kraz_tractor_Chassis::m_body_COM_loc(-2.0, 0, 0.8);
 const ChVector3d Kraz_tractor_Chassis::m_connector_loc(-4.64, 0, 0.82);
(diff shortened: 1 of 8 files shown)


rainergericke (commented):
No problems with configurations and build process. Vehicle behavior seems to be okay. As a test I chose the CRG road course with course and speed controller. This should give us an impression about vehicle behavior in real life situations. It ran without complications.

rserban:
I agree that using a universal joint with yaw and pitch relative degrees of freedom between tractor and trailer is a more appropriate model of a fifth-wheel connector.

However, I question the initialization of the universal joint. Is it supposed to be pitched this much in a horizontal configuration?

Kraz_connector

rainergericke:
I think it is a good start pitch angle, so I don't expect problems. After coupling of tractor and trailor the pitch angle is nearly 0 degrees. A pitched load bed of the semitrailor is highly undesirable. 
Sattelkupplung-Niedrig-Verschiebbar

rainergericke:
In decoupled state there is a small pitch angle, it eases the coupling procedure.

rserban:
Rainer - that may be the case, but is not why there's a pitch angle here. Rather, this is because of a bug in the implementation (specifying an angle in degrees instead of radians).

I will push a fix. While I do this, I can also add an option for a proper pre-set pitch angle.  I'll use a default of 0, but do you have any idea what a good value would be for something like the Kraz?

rainergericke:
Yes, it should be 0, as the drawing suggests.
Bildschirmfoto 2024-10-14 um 16 02 45

rserban:
Thanks, Rainer. Should I then add the option for a pre-set pitch angle at all?

rainergericke:
I think we will not need it.

PR #548, opened 2025-02-22
Add CXX to CSharp demos Cmakelists (not inherited from core cmake)

jdiyn:
For the demo builds for csharp (if chrono was built with openmp or mpi) their find macros require the CXX language tag

Also noted in readme for the demos that user needs to bring irrlicht.dll into build folder manually


jdiyn:
Closing this as latest commit places c# demos into the main solution

PR #552, opened 2025-02-27
fea linear elements refactor

jibril-b-coulibaly:
# This PR refactors the Spring, Bar, and some Beam elements.

This is ongoing work, we should decide when it is a good enough point to merge it

# Springs

### Modifications

- [X] precompute and store rest length as class attribute to avoid computation at each step (this length does not change during simulation)
- [X] refactor force calculation to limit number of length calculations and divisions

### possible issues / bugs to be discussed

- [ ] : the stiffness calculation for the damping does not take the timestep into account. if F = A * v, dF / dx = A / dt ?

# Bars

### Modifications

- [x] precompute and store axial stiffness k=EA/L to avoid computation at each step (this value does not change during simulation)
- [X] refactor force calculation to limit number of length calculations and divisions

### possible issues / bugs to be discussed

- [ ] : the stiffness calculation for the damping does not take the timestep into account. if F = A * v, dF / dx = A / dt ?

# Euler-Bernoulli beams

### Modifications

- [x] define the exact beam reference rotation using the end points and `YDir` vectors in the Builder, rather than using the average Y directions of the end nodes after the fact. This is more practical, but I also believe this was a bug since the average Y direction of the nodes after Gram-Schmidt has no guarantee to be colinear with user provided Y direction of the beams when the nodes pre-exist the beam.
- [x] Refactor the update rotation function: do nothing when corotation is disabled, replace rotation matrix operations with faster quaternion operations
- [x]  Do once and store expensive calculations of the rotation of nodes wrt reference configuration since these never change but were performed on it every step.
- [x] refactor GetStateBlock with lambdas to avoid code duplication
- [x] rename relative rotation quaternions for clarity.

### DIFF
--- a/src/chrono/fea/ChBuilderBeam.cpp
+++ b/src/chrono/fea/ChBuilderBeam.cpp
@@ -94,12 +94,12 @@ void ChBuilderBeamEuler::BuildBeam(std::shared_ptr<ChMesh> mesh,              //
         element->SetNodes(beam_nodes[i - 1], beam_nodes[i]);
 
         ChQuaternion<> elrot = mrot.GetQuaternion();
-        element->SetNodeAreferenceRot(elrot.GetConjugate() * element->GetNodeA()->Frame().GetRot());
-        element->SetNodeBreferenceRot(elrot.GetConjugate() * element->GetNodeB()->Frame().GetRot());
+        element->SetRefRotation(elrot);
+        element->SetElemToNodeArefRot(elrot.GetConjugate() * element->GetNodeA()->Frame().GetRot());
+        element->SetElemToNodeBrefRot(elrot.GetConjugate() * element->GetNodeB()->Frame().GetRot());
 
         element->SetSection(sect);
 
-        element->SetRefRotation(mrot.GetQuaternion());
     }
 }
 
@@ -133,14 +133,12 @@ void ChBuilderBeamEuler::BuildBeam(std::shared_ptr<ChMesh> mesh,              //
         element->SetNodes(beam_nodes[i - 1], beam_nodes[i]);
 
         ChQuaternion<> elrot = mrot.GetQuaternion();
-        element->SetNodeAreferenceRot(elrot.GetConjugate() * element->GetNodeA()->Frame().GetRot());
-        element->SetNodeBreferenceRot(elrot.GetConjugate() * element->GetNodeB()->Frame().GetRot());
-        // std::cout << "Element n." << i << " with rotations:" << std::endl;
-        // std::cout << "   Qa=" << element->GetNodeAreferenceRot() << std::endl;
-        // std::cout << "   Qb=" << element->GetNodeBreferenceRot() << std::endl << std::endl;
+        element->SetRefRotation(elrot);
+        element->SetElemToNodeArefRot(elrot.GetConjugate() * element->GetNodeA()->Frame().GetRot());
+        element->SetElemToNodeBrefRot(elrot.GetConjugate() * element->GetNodeB()->Frame().GetRot());
+
         element->SetSection(sect);
 
-        element->SetRefRotation(mrot.GetQuaternion());
     }
 }
 
--- a/src/chrono/fea/ChBuilderBeam.cpp
+++ b/src/chrono/fea/ChBuilderBeam.cpp
@@ -54,6 +54,8 @@ void ChBuilderBeamEuler::BuildBeam(std::shared_ptr<ChMesh> mesh,              //
         element->SetNodes(beam_nodes[i - 1], beam_nodes[i]);
 
         element->SetSection(sect);
+
+        element->SetRefRotation(mrot.GetQuaternion());
     }
 }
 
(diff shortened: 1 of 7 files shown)


tasora:
Hi - question: in ChElementSpring, you changed 
    ChVector3d int_forceA = dir * internal_force_local;
    ChVector3d int_forceB = -dir * internal_force_local;
into
    ChVector3d int_forceA = dPos * internal_force_local;
    ChVector3d int_forceB = -dPos * internal_force_local;
but I think that the correct form is the original one, where dir is a unit length vector representing direction, whereas dPos is not unit length.
Or have I misunderstood?

tasora:
In ChElementBar, same question...     you compute the two forces via
    ChVector3d int_forceA = dPos * internal_force_local;
    ChVector3d int_forceB = -dPos * internal_force_local;
but it is different from the original
    ChVector3d int_forceA = dir * internal_force_local;
    ChVector3d int_forceB = -dir * internal_force_local;
Right?
A.

jibril-b-coulibaly:
Hi @tasora !

The changes in the force calculation were made to avoid computing the length and scaling multiple times. In the original code below
https://github.com/projectchrono/chrono/blob/19e970e97778ee0064ecf03c6d2049884001353f/src/chrono/fea/ChElementBar.cpp#L112-L114
`GetNormalized` internally computes the length, and then line 114 recomputes the same length. Inside `GetNormalized` the vector gets scaled by `1/length`, i.e., each 3 components are multiplied by `1/length`.

In the refactor below
https://github.com/projectchrono/chrono/blob/f6b06bf9a1ee52fbbbeba838a5642cda76987a10/src/chrono/fea/ChElementBar.cpp#L100-L109

With the non-normalized node-to-node vector `dPos`, we only compute the length once on 102.
the `internal_force_local` in 107 is scaled by `Linv`, so "force" here is a bit of an abuse of language but the math ends up being the same as the original.
The difference is that instead of re-scaling every component of the vector, we rescale the magnitude, doing 1 rescaling instead of 3.

We actually do 2 rescalings, with the damping on 106, so this only saves 1 multiplication by `Linv` compared to the original and is insignificant in terms of performance.
However, once we decide to compute the length only once by not calling `GetNormalized`- which does save compute - this is the more natural way to write the rest of the code in my opinion.

I ran all FEA unit tests and they all pass.

rserban:
@jibril-b-coulibaly Please note that it is **not** enough to run the FEA unit tests to ensure the changes you are making are correct and do not break things. Indeed, most FEA unit tests are for the ANCF formulation. Instead, you should compare results on all appropriate FEA demos.

jibril-b-coulibaly:
@rserban thank you for the information.

I additionally ran `demo_FEA_beamsEuler`, `bin/demo_FEA_beams_constr`  `bin/demo_FEA_beams_rotor`  `bin/demo_FEA_beams_static`  `bin/demo_MOD_slewing_beam` that looks okay.

`bin/demo_FEA_beams_extrude` doesn't look like it is doing much, but that behavior is the same on `main` so it does not seem to be affected by these changes. Looking at the code, when the extruder is updated, a new beam element is created but no orientation (referecnce quaternion) is provided. Is that a bug? This looks like it would be restricted to only work if the beams are aligned with the "World" X,Y,Z coordinates system.
https://github.com/projectchrono/chrono/blob/19e970e97778ee0064ecf03c6d2049884001353f/src/chrono/fea/ChBuilderBeam.cpp#L744-L790

Looking forward, I think the purpose of unit tests is to check that functions operate the way they are supposed to, and for a refactor like the one here, I believe we should be able to rely on them. Does that mean the unit tests coverage is insufficient to test such changes and should be expanded for Euler Bernoulli beams?

If the specific functions refactored have identical behavior as the original per unit testing, there should be no issue. There is always the possibility of functions side effects and I do agree that regression tests are also necessary to see how the changes propagate to the rest of the code.
I found the demos I ran above to mostly/only provide qualitative results. And for those that provide an output, there is no automated comparison to an analytical or "gold" solution, which makes it unpractical to get good comparisons on a large number of demos (run and pipe output for different versions, and diff the files?).
Should we move towards having quantified regression tests (or naming them as such when either current unit tests and demos are actually regression tests), in addition to unit tests and demos? Something that doesn't look off by the naked eye might still be wrong.

Thank you

tasora:
Hi Jibril

Thank you for the refactoring of FEA code.

I had no much time to review all the changes because I am quite busy these days. But all looks correct so far.

Just keep in mind that many finite elements in Chrono were implemented many years ago, and they really need a refresh. For example I was planning to rewrite the solid elements (tetrahedrons etc. ) both to use a total lagrangian formulation and to accommodate a new design that uncouples the pde from the element, in the sense that one could apply both elasticity and thermal problems to the same element for example. I am doing something along this line in the peridynamics/ branch.
Etc etc.

Thanks

Alessandro


________________________________

jibril-b-coulibaly:
Hi everyone.

I think this PR is almost good to go.

I will transfer the points I noted above about Gram-Schmidt to a new issue / PR.

The only outstanding question I have is about the damping force in the stiffness matrix.
I do think the stiffness matrix should

-  either take the timestep into account (i.e., if F = A * v, dF / dx = A / dt)
- or not take viscous forces into account at all and simply be the elastic stiffness.

Please let me know your thought on this last point. I am also fine shelving it as an issue for now.

Thank you,
Jibril

PR #556, opened 2025-03-17
Fix the CI/CD Pipeline on Mac by using absolute path to find blaze

uebian:


uebian:
> Could you please change CMakePresets.json for macosci so that it's consistent throughout: use either ~/dependencies/... or $env{HOME}/dependencies everywhere.

The issue has been addressed in https://github.com/projectchrono/chrono/pull/568

uebian:
>  Could you please look into enabling that module?

The issue has been addressed in #571

PR #564, opened 2025-04-11
Add Viper Direct Control Driver

JasperGrant:
In my own research I have found it very useful to have individual unrestricted control of all four wheel speeds, steering angles, and suspension angles for the Viper rover.

For instance there is not currently a way to move individual wheels in opposite directions in the current two drivers.

Happy to make changes to the way this is implemented if required!


Huzaifg:
This looks great and useful. Thanks for the contribution

Issue #566, opened 2025-04-18
Option to setup direct solver once for entire simulation and re-use matrices assembly and factorization


jibril-b-coulibaly:
## Problem Statement / Feature Request

I have been profiling FEA simulations that use HHT timestepper and PardisoMKL solver (see kcachegrind image below).
I noticed the `ChDirectSolverLS::Setup()` function, in particular the factorization `ChSolverPardisoMKL::FactorizeMatrix()` takes up a very large majority of the computational time.

The system I am simulating is subject to slow and small transformations, so that the stiffness and mass matrix is nearly constant throughout the simulation. Therefore, a lot of time is wasted factorizing a matrix that is nearly constant.

This simulation already uses the `modified_newton` flag in the HHT timestepper, which calls Setup once per step (called 4x in the image below) and reuses the factorization for each iteration (called 13x in the image below).
My proposal/request is to implement an additional flag that only calls setup **once at the beginning of the simulation**, and stores persistent initial matrices, and re-uses them throughout the entire simulation.

The tradeoff for using initial matrices can be higher memory costs and a smaller convergence rate / larger number of iterations, but given the high cost of factorization I believe this would be beneficial for FEA simulations where these matrices are almost constant (small transformations), or where using the initial elastic stiffness is commonplace (e.g. fracture mechanics simulations where softening and loss of positive definiteness can cause numerical difficulties).

## Implementation ideas

I create a draft pull request #567 so we can collaborate on implementing the feature. Here are some ideas so far:

- Implementing a new flag at the timestepper level, similar to the `modified_newton` flag, seems to be the simpler approach because of the similarities with existing capabilities..
- We can start with HHT timestepper and once we have it tested and working, generalize it to all timesteppers.
- If this is Solver-agnostic, it could work for `MINRES` as well. If not we might start with Pardiso and later consider how to interface various solvers.

![Image](https://github.com/user-attachments/assets/dd2ac441-b858-4f9f-9a2b-9845478054ca)

rserban:
@jibril-b-coulibaly  - yes, this feature is planned, but will be implemented in a more robust manner than what you suggest here.

jibril-b-coulibaly:
@rserban 

Thank you for a prompt response.

I should have brought more context to this issue.
This is a feature I have been discussing with @tasora @dariofusai93 @dariomangoni over the past month as it is a big bottleneck in some of the work we do.

We met today and decided making an issue and PR would be the best way to move forward in a collaborative way.

[...]

PR #567, opened 2025-04-18
Implement matrix assembly and factorization persistent through entire simulation

jibril-b-coulibaly:
Assemble and factorize direct solver matrices only once at the beginning of the first step

corresponding issue: #566

### DIFF
--- a/src/chrono/timestepper/ChTimestepperHHT.cpp
+++ b/src/chrono/timestepper/ChTimestepperHHT.cpp
@@ -34,7 +34,8 @@ ChTimestepperHHT::ChTimestepperHHT(ChIntegrableIIorder* intgr)
       h_min(1e-10),
       h(1e6),
       num_successful_steps(0),
-      modified_Newton(true) {
+      modified_Newton(true),
+      persistent_Newton(false) {
     SetAlpha(-0.2);  // default: some dissipation
 }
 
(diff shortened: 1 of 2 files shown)


rserban:
I *strongly* suggest you ask us first before starting work on a new PR.  
In particular, this type of feature has been on my todo list for a while and I intend to implement it in the near future.  Having said that, I will implement something more robust and general than what you proposed, where the Jacobian is evaluated and factorized when needed and reused over multiple steps as long as it's still appropriate.

jibril-b-coulibaly:
@rserban as I clarified in #566 , this is something I discussed with @tasora @dariofusai93 @dariomangoni .

I thought they might be in charge of triage for this PR/issue, and that maybe you had talked about it without me as well.

Let's find a way forward together

jibril-b-coulibaly:
@rserban following our discussions, I pushed an update that allows multiple options for `modified_Newton` methods, changing the boolean flag to an `enum` type named `ModifiedNewton`.


- `ModifiedNewton::UNMODIFIED` corresponds to classical Newton (former `false` boolean value)
- `ModifiedNewton::EVERY_STEP` corresponds to updating the Jacobian very step (former `true` boolean value)
- `ModifiedNewton::CONSTANT` corresponds to re-using the first Jacobian computed across all steps and iterations (feature proposed in the PR)
- `ModifiedNewton::AUTOMATIC` is reserved for implementing the feature you are working on where the Jacobian is computed automatically when needed.

I have generalized the `call_setup` logic and placed it in a dedicated function called at each iteration to determine if setup should be called. This is where the `ModifiedNewton::AUTOMATIC` option can be implemented. This change does not modify existing functionality and replicates the current behavior.

PR #580, opened 2025-06-06
Add color setting for the Viper Rover

JasperGrant:
I noticed that while individual viper parts had colors assigned they were never applied to the rover.

If it is intended for these colors to not be applied I can definitely delete this PR.

Thank you!

### DIFF
--- a/src/demos/fmi/cosim/fmu2_actuator/actuatorFMU.cpp
+++ b/src/demos/fmi/cosim/fmu2_actuator/actuatorFMU.cpp
@@ -161,8 +161,6 @@ fmi2Status FmuComponent::exitInitializationModeIMPL() {
     integrator->SetMaxIters(50);
     integrator->SetAbsTolerances(1e-4, 1e2);
 
-    sys.DoAssembly(AssemblyAnalysis::Level::FULL);
-
     return fmi2Status::fmi2OK;
 }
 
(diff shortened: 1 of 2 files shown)


rserban:
Thanks for looking into this.

It looks like this model was not implemented in the best way:
- using a ChVisualShapeTriangleMesh here is overkill (and "repairing" a mesh meant for visualization only even more so)
- using this type of visual shape is also limiting (e.g., if one also has an associated Wavefront .mat file)
- for visualization using meshes, a much better option is using ChVisualShapeModelFile as that will automatically use a materials file if one exists (which is the proper way of providing textures and colors for such visual assets)

So, if you're interested in making some fixes, here's what I suggest:
- include `ChVisualShapeModelFile.h` instead of `ChVisualShapeTriangleMesh.h`
- replace the code in the `if(m_visualize)` block with something like:
```     
        auto vis_mesh_file = GetChronoDataFile("robot/viper/obj/" + m_mesh_name + ".obj");
        auto shape = chrono_types::make_shared<ChVisualShapeModelFile>(vis_mesh_file);
        shape->SetMutable(false);
        m_body->AddVisualShape(shape, m_mesh_xform);
```
- get rid of the `ViperPart::m_color` member variable altogether

By the way, the same issue is present in the Curiosity rover model, so maybe modify both?
Let me know if you intend on submitting a new PR with the fixes above. Otherwise, I will push them myself.

JasperGrant:
Hello Dr. Serban,

Happy to put some work in on this! I have added the changes you recommended to the Viper.cpp and Curiosity.cpp files to this PR and have gotten some weird visual effects. (see below)

image

image

I did notice there is not material file in the data/robot/viper or curiosity folders. Could this be why we are seeing odd gaps in the wheels and white/black colors?

Thank you,
Jasper Grant

PR #583, opened 2025-06-24
Constitutive eqn 

AkpanWisdom:
updated the constitutive equations and added the random_field


AkpanWisdom:
Sorry, this was meant for a different fork

PR #633, opened 2025-11-04
VSG speed optimisations

jdiyn:
Hi all, with the VSG module, I made some Render loop optimisations after running some profiling - which bring notable speedup in c++, c# and python vsg demos.

I found a significant bottleneck was indiscriminately binding the rendering pipeline to the simulation step rather than decoupling within the VSG class. So it was running 3-400 fps (trying to keep up with the time-step) and transferring cpu to gpu chewed 30-40% of total cpu time.
Python users might forget to separate the rendering from the simulation steps in the code; since Irrlicht was not necessarily rendering every step, but VSG was attempting to copy step-changes across to GPU every sim step unless the app/demo code explicitly conditioned out the vsg scene render. Thought a set-target-fps function might be more intuitive... if not set, it is init to 0, which disables the render-gate as if it weren't there.

I also discovered ChVisualShapeSpring::UpdateLineGeometry() was being called to rebuild CPU-side geometry every physics step even though VSG can generate coils procedurally on the GPU (the cpu build was a memory hog in some profiled snapshots). This was causing a dirty() flag to require a scene traversal every time and rebuild from the cpu geometry. I added an opt-in flag that VSG sets during binding to skip this redundant work while preserving backward compatibility if users want it

Finally, DearImGui was giving a message for the vehicle menu sliders (throttle, steering, braking) because they didn't have distinct id's, which could cause mem/perf problems. Pushing through distinct id names solves this.

There's probably many more optimisations to be had with the vsg module, but I thought these might be a 'quick win' which boost the perf without much impact operationally.

Let me know what you think 
-josh

### DIFF
--- a/src/demos/python/vehicle/demo_VEH_CRMTerrain_WheeledVehicle.py
+++ b/src/demos/python/vehicle/demo_VEH_CRMTerrain_WheeledVehicle.py
@@ -70,6 +70,10 @@ visualization_sph = True
 visualization_bndry_bce = False
 visualization_rigid_bce = False
 
+# Output settings
+save_images = False  # Set to True to save BMP images
+save_results = False  # Set to True to save results.txt
+
 # CRM material properties
 density = 1700
 cohesion = 5e3
@@ -261,9 +265,12 @@ os.makedirs(out_dir, exist_ok=True)
 out_file = os.path.join(out_dir, "results.txt")
 
 # Use a simple CSV writer
-import csv
-csvfile = open(out_file, 'w', newline='')
-csvwriter = csv.writer(csvfile, delimiter=' ')
+csvfile = None
+csvwriter = None
+if save_results:
+    import csv
+    csvfile = open(out_file, 'w', newline='')
+    csvwriter = csv.writer(csvfile, delimiter=' ')
 
 # -----------------------------
 # Create run-time visualization
@@ -290,6 +297,8 @@ if render:
     visVSG.SetCameraAngleDeg(40)
     visVSG.SetChaseCamera(chrono.VNULL, 6.0, 2.0)
     visVSG.SetChaseCameraPosition(chrono.ChVector3d(0, 8, 1.5))
+    # By default the target render FPS is 60 for all vehicle VSG visualisations, set to 0 to disable
+    # visVSG.SetTargetRenderFPS(200)
     visVSG.Initialize()
     vis = visVSG
 
@@ -329,6 +338,8 @@ while time < tend:
         vis.Render()
         render_frame += 1
 
+
+
     # Synchronize systems
     driver.Synchronize(time)
     if vis:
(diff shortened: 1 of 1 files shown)


rserban:
This is fantastic, Josh!

Did you by any chance also profile some SPH simulations with the FSI module (or CRM deformable terrain, as that uses the same SPH solver and visualization)?

jdiyn:
Thanks Radu,
I did not profile the SPH yet, but I did some just now on the demo SCM Terrain wheeled vehicle. I found that we're copying the entire dynamic mesh each render() even when its unchanged - that would need some caching and rework to improve significantly, such as maybe storing the mesh on the gpu and only updating the altered vertices each render. But that will need tracking and some testing. 

In the mean time, I ran a quick test on improving what's currently there in the loop for dynamic meshes:

image

we're loop copying vertices with temporary vec's every frame.

I added a new commit to this PR, which profiled ->
image

I haven't testing it where there's dynamic normals, but used the same method in that block as in the vertices/colours blocks.

I think we could probably boost this vsg handling down to miminal cpu% load with some more significant shifts in tracking altered data, caching the unchanged on the gpu, and marking dirty but only trasnferring the changed data. But that would need some testing to intergrate well. Additionally, could consider a compute shader with receives only the altered vertices and runs the rest on the gpu and transfers results over to the render queue - but that would also be quite a bit of dev i think.

I'll add the FSI module to my build and do some tests later today

rserban:
Thanks, Josh. 

The SCM code includes functionality to report only grid nodes that were modified during the last time step. If I recall correctly, that's used in the Irrlicht module, but not VSG. 

Please let me know what you find with SPH visualization. There, we may need to branch (if needed) between fluid simulations (i.e., when solving Navier-Stokes) vs CRM (deformable terrain) simulations. In the latter case, we can easily identify particles that were changed during the last time step. In the former, all particles are affected all the time.

In any case, I suspect there are other, lower hanging fruit, optimization that can be done in VSG for particle clouds. Just like you did above.

jdiyn:
Hi Radu, apologies, it took me some time to get the build going on my laptop (for the cuda requirements), I've found a similar cpu hog in the temporary vec3 and vec4, similar to before. 

image

However, in simply copying the same approach as for dynamic mesh over, there's a speed up, but it revealed some cpu hogging from pulling all the colours for every particle (and also im a bit uneasy about using raw approach above that may work with dynamic mesh becuase the particle count doesn't change, but could be unsafe/undefined with varying particle count sims. I'm looking to see if there's an easy solution there and will add the change here if its successful

jdiyn:
Hi Radu,

I've grabbed another lowish hanging fruit and bundled the particle cloud dynamic colouring loops into a bulk transfer to a small vulkan compute shader (had to extend chronovsg to insert the compute shader into the pipe)

pulled the time spent on the cpu for chrono vsg in fsi and CRM vehicle demo down to small single digit percentages on my system.

I have not yet tested the above on the python vsg approaches.

jdiyn:
I've tested with Python, and with the CRM wheeled vehicle demo - while my laptop isn't particularly fast (~5yrs old) - I get an RTF of about 13-14 in c++
In Python, I get an RTF of ~17-18 on the same demo.

Please note I didnt have the opportunity to built the fsi module with using_double_on_sph, so it's all single-based results here.

rserban:
This is really nice, Josh! I'm very eager to try this out as soon as I return from travel on Friday.

Regarding precision, we use single precision almost exclusively in Chrono::SPH.

PR #634, opened 2025-11-05
Fix default values and documentation for ChContactMaterial

skvhk:
Hi,
just two very minor fixes to ensure that documentation and code match. I wasted some time wondering about the restitution. Documentation said default==0 but it had been set to 0.4 by default.
Also added the default value for sliding friction in the doc.

### DIFF
--- a/src/chrono/physics/ChContactMaterial.h
+++ b/src/chrono/physics/ChContactMaterial.h
@@ -51,7 +51,7 @@ class ChApi ChContactMaterial {
     /// Get the static friction coefficient.
     float GetStaticFriction() const { return static_friction; }
 
-    /// Set the sliding (kinetic) friction coefficient.
+    /// Set the sliding (kinetic) friction coefficient (default: 0.6).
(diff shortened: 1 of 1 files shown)


rserban:
Thanks for pointing out these inconsistencies.
In general, I am not a big fan of PRs that only change some small thing because they are often incomplete.  I'll merge this one time, once you make a few changes:

- Unless you have checked that no Chrono demo or test is affected by the change of the default value for the coefficient of restitution, the correct way to resolve an inconsistency between implementation and documentation is to change the documentation.  This is because there may be code out there that relies on the current default value.
As such, please leave the default value at 0.6 and fix the function documentation.
- While you are at it, why not fix documentation of other functions in ChContactMaterial.h by providing the default value in a consistent manner? E.g., SetStaticFriction
- There are also some typos that should be fixed, including in the comment for the function you touched in the header file ("ormal" instead of "normal").  Might as well fix all these minor issues in a single commit.

Thanks

PR #636, opened 2025-11-14
chLinkNodeFrame fix load residual Cq'*L

Yohanumerics:
I think there might have been some copy-paste issues in this function, based on the other functions of the chLinkNodeFrameGeneric class


rserban:
Thanks for spotting this one as well!
Yet another class not used in any demos or tests, hence why this bug survived for so long.

PR #637, opened 2025-11-14
Fix operator *= on ChStateDelta

Yohanumerics:
ChStateDelta's operator *= is overridden, and uses the operator * of its base class (instead of *=), making a statement such as "Dv *= c;" not doing anything on the Dv vector.


rserban:
Thanks for spotting this. 
Fortunately, this operator was used only in one place in the Chrono core library, where it did not have important effects.

PR #648, opened 2025-11-27
managed DLL for csharp swig emitted .cs files

jdiyn:
Hi,
Submitting a PR to introduce an optional managed DLL (`chrono_csharp_wrapper.dll`) for the Chrono C# module swig generated .cs files

effectively its set up to  streamline downstream projects and demos (rather than embedding all .cs files into the exe; we can just use the .dll along with the other chrono dlls)

When enabled with the cmake option, all SWIG-generated C# sources are bundled into a single DLL and set as the target for the csharp demos

Potentially, this could be renamed from 'wrapper' - perhaps to chrono_csharp_managed - if the meaning/purpose of the dll is unclear

thanks!
-josh


uebian:
If this option is tested on Linux, if not if it's expected to work on Linux? I might need to change CI/CD code accordingly

jdiyn:
I haven't tested on linux, but i did try to write it to be generic for linux & macos. if you could test on those systems I would greatly appreciate it!
-josh

PR #652, opened 2025-12-10
[FSI] Fix incorrect torque readout in demo_FSI_SingleWheelTest

becarpg13:
While using the Chrono::FSI module for terramechanics simulations, specifically for rover wheel analysis, I noticed that the torque output in the demo was yielding inconsistent values, close to zero, at the order of 1e-26. The issue stems from the output logic querying the torque from the linear actuator rather than the angular motor.

**Changes**
* Updated the data retrieval logic to query the motor torque from `ChLinkMotorRotationAngle` rather than `ChLinkLockLinActuator`
* The motor definition was moved out of the `CreateSolidPhase` function, so it can be called at main to output

Happy to contribute!

Best,
Rebeca Guimarães

### DIFF
--- a/src/demos/fsi/demo_FSI_SingleWheelTest.cpp
+++ b/src/demos/fsi/demo_FSI_SingleWheelTest.cpp
@@ -75,6 +75,7 @@ double dT = 2.5e-4;
 
 // linear actuator and angular actuator
 auto actuator = chrono_types::make_shared<ChLinkLockLinActuator>();
+auto motor = chrono_types::make_shared<ChLinkMotorRotationAngle>();
 
 // Save data as csv files to see the results off-line using Paraview
 bool output = true;
@@ -246,7 +247,6 @@ void CreateSolidPhase(ChFsiSystemSPH& sysFSI) {
     sysMBS.AddLink(prismatic2);
 
     // Connect the wheel to the axle through a motor.
-    auto motor = chrono_types::make_shared<ChLinkMotorRotationAngle>();
     motor->SetName("engine_wheel_axle");
     motor->Initialize(wheel, axle, ChFrame<>(wheel->GetPos(), chrono::QuatFromAngleX(-CH_PI_2)));
     motor->SetAngleFunction(chrono_types::make_shared<ChFunctionRamp>(0, wheel_AngVel));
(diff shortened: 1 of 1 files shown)


rserban:
Hi Rebeca - thank you for the PR. You are absolutely right, this was a bug and your fix is correct.

Having said that, I was thinking of removing this particular demo.  That is because there is a much better way of setting up such a single wheel test and that is through the utility class [ChTireTestRig](https://github.com/projectchrono/chrono/blob/main/src/chrono_vehicle/wheeled_vehicle/test_rig/ChTireTestRig.h).  In particular, [demo_VEH_TireTestRig_CRM](https://github.com/projectchrono/chrono/blob/main/src/demos/vehicle/test_rigs/demo_VEH_TireTestRig_CRM.cpp) shows usage of this test rig on CRM deformable terrain.  ChTireTestRig offers a lot more capabilities than the hard-coded mechanism in demo_FSI_SingleWheelTest.

Could you please see if this alternative works for your needs? Unless I see a compelling reason for keeping it around,f or the sake of eliminating unnecessary duplication, I intend to remove demo_FSI_SingleWheelTest in the near future.

--Radu

becarpg13:
Hi Radu,

That sounds perfect. Please feel free to remove the demo.

The primary reason I was using the `demo_FSI_SingleWheelTest` was that I have been working with a local copy of the repository to maintain stability during my undergraduate thesis, so I hadn't caught up with the latest updates until now. Additionally, my thesis committee specifically requested a demonstration of the explicit low-level setup, which the demo code pipeline illustrated quite clearly.

I just tested the new rig setup and it is excellent! The visual improvements are a huge plus. I also really appreciate the new towed/driven options and especially the settling time feature. 

I actually tried to implement a settling phase myself right before the defense, but I hit a dead end trying to ramp up the actuator functions (attempting something like `wheel_vel*constant*dT` inside `SetActuatorFunction` and reassigning it), which ultimately didn't work. I had major issues with the wheel dropping too fast and scattering particles everywhere right at the start, so having this natively implemented is a lifesaver and will save me weeks of coding in the next semester.

Regarding performance, I noticed the new rig demands a bit more from the hardware (slightly higher RTF), but the trade-off is absolutely worth it. It feels like a true digital twin of the physical test rigs.

I see no reason to keep the old demo if the intention is to streamline the codebase. I have my local backups, and as I recall, a version of that implementation can still be found in the repository for the gravitational offset paper, so it remains accessible if anyone specifically needs that setup.

Thanks again for the continuous improvements!

-- Rebeca

rserban:
Hi Rebeca,

Thank you for the background information. 

Your work sounds very interesting and I'd be very interested in learning more about what you do. At the very least, please send me a copy or your thesis when done. Of course, not on GitHub :)

Let me know if you find any missing features in the test rig (or anywhere else in Chrono) or simply provide a PR if you feel so inclined.

Finally, you are right, nothing is ever lost in git. Even after I remove that redundant demo, you'll be able to find it at previous commits.

Best,
Radu

PR #665, opened 2026-01-06
add collisionSystem to make the example work

TomMontagnon:
Without explicitly setting the collision system, the cube passes through objects, which makes the example confusing.

It might be helpful to add a short explanation in the code or the documentation mentioning that gravity is already set by default on the Y axis.


rserban:
Thanks for spotting this issue!, Tom

I will merge this PR, but only a couple of comments for the future:
- something like this does not really require a PR; opening an issue would have been arguably more appropriate.
- do not conflate different issues in the same place; your observation about the gravity setting is certainly pertinent, but should have been a separate GitHub issue. Having said that, I will make the necessary updates shortly.

PR #700, opened 2026-03-14
Align member names in Python demo with C++ code

uebian:


uebian:
Pushed directly in main, closing the PR

PR #731, opened 2026-04-25
AMD HIP port compability initial

fl0rianr:
This is a port of the Chrono code to be AMD HIP compatible. The CUDA parts have been modified to support a dual CUDA-HIP backend. This affects 34 files, with what I consider minimal and useful changes.

Three modules are affected:

Chrono DEM: fully ported
Chrono FSI SPH: fully ported
Chrono Sensor: runs without OptiX on AMD, but the OptiX part is not ported

Runtime checks were performed, but only on AMD gfx1151 (Radeon 8060S iGPU); therefore, testing was **not** very extensive.

CUDA still runs fine as well, tested on an RTX 4090.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -313,26 +313,151 @@ if(MPI_CXX_FOUND)
 endif()
 
 #-----------------------------------------------------------------------------
-# CUDA support
+# GPU backend support (CUDA and HIP)
 #-----------------------------------------------------------------------------
 
-message(STATUS "Searching for CUDA")
-
-# Notes:
-# - with new policy CM0104, CMake sets CMAKE_CUDA_ARCHITECTURES based on what nvcc reports
-# - nvcc uses as default sm_52
-# - we use our own cached CHRONO_CUDA_ARCHITECTURES variable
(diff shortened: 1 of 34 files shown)


fl0rianr:
@rserban As discussed, here is my PR. Please let me know what you think.

rserban:
Thank you, @fl0rianr!

This looks good. I will merge the PR shortly.

One comment:  Keeping this set of changes minimal and preventing changes to source files made perfect sense for now. But at some point, I think we'll refactor some of the function names to replace the `cuda` prefix with something like `gpu` (for things defined in the new `ChGpuRuntime.h` and `ChSphGpuRuntime.h` files) for clarity.  I'll take care of this at some point (unless you see some potential issues with such a renaming; I cannot think of any).

Also, we will need to provide some documentation on installing Chrono::DEM and Chrono::FSI-SPH on AMD HIP platforms. I hope we can work together on doing this in the not very far future.

fl0rianr:
Regarding the function names, that's the right call. But I want to add an AMD performance fix before you do so. Thanks for being patient.

rserban:
Sure thing. There's no hurry

PR #734, opened 2026-04-27
minimal cuda 13 compilation fix

fl0rianr:
This is a simple CUDA 13 fix not related to changes made in the bigger PR #731 

(worked in a dev-container with older cuda and compiled without open the container now and saw this issue)

### DIFF
--- a/src/chrono_dem/cuda/ChDem_SMC.cuh
+++ b/src/chrono_dem/cuda/ChDem_SMC.cuh
@@ -24,6 +24,7 @@
 #else
     #include <cub/cub.cuh>
 #endif
+#include <thrust/functional.h>
 #include <cassert>
 #include <cstdio>
 #include <fstream>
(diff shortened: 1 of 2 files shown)


rserban:
Thanks, Florian. 
We are not supporting CUDA 13 officially yet, but this is obviously on the todo list. I have not tried building the CUDA modules with version 13.0 yet. Does this mean that Chrono::FSI-SPH builds and works fine with CUDA 13?

fl0rianr:
> Does this mean that Chrono::FSI-SPH builds and works fine with CUDA 13?

Yes, indeed.

On a side note: For CUDA 14 double4 is depreciated. Just to let you know far in advance.

rserban:
Excellent. 
Then, one more quick question: did you by any chance also build Chrono::Sensor with OptiX support using CUDA 13? Does that compile and work fine?

fl0rianr:
Compilation works for Chrono::Sensor as well. Since I'm not to deep into this huge project I actually don't really know how to test the runtime of that... does that look  alright for you?
```bash
chrono_contribution/build/bin$ ./demo_SEN_GPSIMU
Copyright (c) 2019 projectchrono.org
Chrono version: 10.0.0


Simulation time: 20s, wall time: 35.1923s.
´´´

uebian:
Thank you for contribution. `demo_SEN_GPSIMU` doesn't involve in OptiX visualization. If possible, could you please run `./demo_SEN_camera` and provide a screenshot?

fl0rianr:
Sadly it does not work in the current state, so i created PR #735  if that's fine with you.

PR #735, opened 2026-04-28
Fix sensor otpix includes and NVRTC CUDA 13

fl0rianr:
The sensor module is now functional with OptiX.

Include directories have been updated for CUDA 13.

The most critical change is the requirement of **C++17** for the NVRTC compiler. Since this is a development build and Chrono 11 is unlikely to be used with legacy build tools or older operating systems (older such as Debian 9), adding this flag specifically for the RTC stack is not problematic.
Everything works fine (note: the top two windows were stretched).

image


rserban:
Thanks (again)!

By the way, we have decided to drop support for C++14 throughout and expect at least C++17.  This change will be made in the near future.

PR #737, opened 2026-04-30
Improve macOS CMake configuration and build scripts

vlordier:
## Summary
- Enable `CMAKE_MACOSX_RPATH` on macOS so binaries find their dependent `.dylib` files in the build tree without requiring a full install
- Remove stale Xcode `GCC_VERSION` attribute that references a compiler unused since Xcode 4.x
- Remove dead `MAC_LIBS` variable from Chrono::Irrlicht linking (was never defined; `Irrlicht::Irrlicht` carries its own framework dependencies)
- Delete unused `cmake/FindModernCXX.cmake` (never invoked by any CMakeLists; C++ standard detection duplicates its logic inline in `src/CMakeLists.txt`)
- Support `CHRONO_DATA_DIR` environment variable for Chrono data path resolution (fixes VSG font/asset loading from arbitrary working directories)
- Modernize `contrib/build-scripts/macos/buildChronoMac.sh`

## Build script improvements
- Auto-detect `arm64` vs `x86_64` architecture at runtime
- Replace hardcoded Homebrew paths (e.g. `libomp/15.0.7`) with `brew --prefix` for version independence
- Add pre-flight dependency checks for `cmake`, `swig`, and `python3` with clear error messages
- Version-independent OpenMP detection (`brew --prefix libomp`) with fallback messaging
- Version-independent Python detection using the correct Homebrew Python binary, with include/library path validation
- Guard Homebrew-dependent paths (`EIGEN3_INSTALL_DIR`, `IRRLICHT_ROOT`) behind non-empty `HOMEBREW_PREFIX`
- Only pass `-G` to CMake when a build system is detected (avoids empty generator argument)
- Rename `LIPOMP_*` to `LIBOMP_*` for clarity
- Shellcheck-clean (0 warnings)

## Code review fixes
- Guard `CMAKE_MACOSX_RPATH ON` with `AND NOT DEFINED` to respect user overrides
- Fix Python version detection to use Brew Python instead of system `python3`
- Validate Python include/library paths before passing to CMake

## Verification
- CMake configure and build pass cleanly on macOS arm64 (Apple Silicon, macOS 26.3.1, AppleClang 21)
- All 8 Chrono::VSG demos launch successfully via Vulkan/MoltenVK
- All 4 Chrono::Irrlicht demos launch successfully
- 80+ core/FEA/MBS demos build and run

## Backward compatibility
All changes are macOS-specific. Linux/Windows builds are unaffected.


vlordier:
@copilot apply changes based on the comments in [this thread](https://github.com/projectchrono/chrono/pull/737#pullrequestreview-4206736560)

rainergericke (commented):
Hi, I am unable to start demos from the Xcode IDE, because the data path to the chrono resources is unknown. If the CHRONO_DATA_DIR is set in Xcode, there is no effect. That makes the Xcode IDE useless. What did you make it run on your Mac?

rserban:
@vlordier - this PR is getting a bit out of hand and conflates too many different things.
It is unlikely we will want to merge it in this form.

Please submit PRs focused in one unified set of changes, so that we can properly evaluate them individually.

Please also answer questions posed in the conversation. I don't think the question by @rainergericke got an answer from you.

Finally, please follow the naming style used in Chrono (e g., for the demo you added).

Thanks!

vlordier:
**Root cause identified**: The arm64 SCM hang was caused by out-of-bounds vector access in `ChBody::accumulators`. On x86 this silently corrupted memory; on arm64's stricter memory model it hangs.

**Essential fix** (commit 8ff8ae5):
`ChBody::EmptyAccumulator`, `AccumulateForce`, `AccumulateTorque` now check `idx >= accumulators.size()` and resize before accessing. This is a one-line fix per method.

**Safety improvements** (included on this branch):
- `#ifndef __APPLE__` guards on OpenMP pragmas in SCMTerrain.cpp (prevents potential arm64 OpenMP issues)
- Flood-fill iteration guard (prevents potential infinite loops)
- `BT_DISABLE_STACK_TEMP_MEMORY` for Bullet ray test (alignment safety)

**Verified**: SCM terrain + body contact works on arm64. `AccumulateForce` can be called before first `DoStepDynamics`.

uebian:
@vlordier I noticed that you pushed a fix to solve the SCM hanging issue on macOS, could you please share how to reproduce the SCM hanging issue on macOS (for example, by running which demo)?

vlordier:
@rserban @uebian I will close this one and open a tighter one indeed.

rserban:
That sounds great, @vlordier.

Please create multiple PRs, for each separate issue.  For instance, while I assume you found the problem with the SCM implementation on MacOS, that is a stand-alone issue. Create a PR only for that fix. I don't have access to a mac, so I wouldn't be able to assess things specific to that platform (others will look at that). But I can look at the SCM fix and merge that PR.

Unless you already do so, please also apply the clang-format that we use throughout. If you add new demos, please follow the naming convention we use (demo_[***]_name, where [***] is in all caps and indicates the Chrono module containing the feature that the demo is primarily about ) .

PR #745, opened 2026-05-30
Replace Blaze with Eigen in chrono_multicore

IanRPage:
This PR completes the migration of the chrono_multicore module from the Blaze linear algebra library to Eigen, eliminating Blaze as a third-party dependency.

**Changes**

**Dependency removal**

- Removed cmake/FindBlaze.cmake, Blaze build scripts, and all blaze_DIR / blaze_INCLUDE_DIR references from CMake and docs
- Deleted fast_matrix_market/app/Blaze.hpp and Blaze-specific unit test (test_blaze.cpp)

**Core rewrite**
- Replaced all DynamicVector<real> and CompressedMatrix<real> types with VectorType and SparseMatrixType
  (Eigen aliases) across ChDataManager, solvers, and constraints
- Rewrote ComputeInvMassMatrix to use Eigen's reserve()-per-row sparse construction, avoiding repeated
  reallocation
- Rewrote ChSchurProduct: added a Setup() method that caches block submatrix views (middleRows/middleCols)
  ahead of the solve loop, and applied .noalias() throughout to eliminate temporaries

**Correctness fixes**
- Fixed ascending column order for off-diagonal inertia entries in ComputeInvMassMatrix
- Fixed operator precedence bug in the GS solver
- Fixed columnwise product in ChSchurProduct
- Fixed element access in Jacobian matrices

**Other**
- Reduced redundant calls in the iterative solver main loop
- Replaced scalar loops with vectorized Eigen expressions (e.g., segment().setZero(), cwiseProduct)
- General cleanup of temporary variables and unused #pragma omp parallel for annotations where incorrect


rserban:
Ian - everything looks good, but there are a few small things that require changes:

1. Remove `CMakeFiles/cmake.check_cache` and `CMakeCache.txt` from committed files.
2. Undo changes to `CHANGELOG.md` – the mentions to Blaze in that file must remain as they pertain to previous releases.
3. Undo all changes to files in the `src/chrono_thirdparty` directory. You removed a file from `fast_matrix_market` and modified a file in `yaml-cpp` (the latter seems to be just reformatting).  We do not touch the packages in src/chrono_thirdparty – these are written by others and we use them as-is.
4. Remove the use of `__builtin_prefetch` in ChSystemMulticore.cpp.  That does not work on Windows (and so, if we were to keep it, we’d need to use a conditional and use the MSVC version, `_mm_prefetch`). But then, I don’t know what other compilers might need. Furthermore, I don't expect prefetching to add significant performance here.

Thanks!
--Radu

IanRPage:
@rserban thanks for catching that, got it fixed

PR #751, opened 2026-06-08
fix: cmake error in specific sensor module

fl0rianr:
**Why:** 

During the linking of demo_SEN_CRM_Rendering an error occurs while CH_ENABLE_MODULE_VEHICLE is not set.

**Changes:**

Apply a warning in this specific build config and avoid building demo_SEN_CRM_Rendering without necessary dependencies.


rserban:
Thanks Florian, but I will close this PR. It's not clear why we would ever need to link Chrono_vehicle in the sensor library.  The fix to the CMake configuration for the sensor demos is correct, but I opted to add that fix in a separate commit (to include a couple more things for cleaning up that CMakeLists.txt file.

Issue #507, opened 2024-07-12
Serialization of a std::map<std::string, ChLinkTSDA> leads to a segmentation fault


paulotto:
Hi everyone,

I tried to serialize a class member of type *std::map<std::string, ChLinkTSDA>* with *archive_out << CHNVP(map_);*, but unfortunately, this led to a segmentation fault.

While looking for the cause of this, I came across another issue, which I wondered if was related to the first issue. If you want to do something like this
```cpp
std::map<std::string, ChLinkTSDA> map_{
    {"l1", ChLinkTSDA()}
};
```

a 'bad_alloc' exception is thrown. I could eliminate this by adding the following line to the copy constructor of ChLinkTSDA:
```cpp
m_jacobians = other.m_jacobians ? new SpringJacobians(*other.m_jacobians) : nullptr;
```

But in the end, I couldn't solve the serialization problem. Maybe you have a solution.

Thanks and best regards,
Paul

DanNegrut:
Thanks for the detailed report, and I am sorry it sat this long without a reply.

Both halves of what you found were real. One was fixed shortly after you filed, and the other was
never fixed and still reproduces today, so this issue stays open.

### The `bad_alloc` is fixed

Your diagnosis was correct, and so was your patch. The old copy constructor never assigned
`m_jacobians`, and never assigned `m_variables` on the path where the source had no ODE registered,
while `~ChLinkTSDA()` deletes both. A default-constructed `ChLinkTSDA()` copied into a `std::map`
therefore handed two uninitialized pointers to `delete`, which corrupts the allocator's bookkeeping and
surfaces as a failure at some later, unrelated allocation.

That was fixed in 85725a0277 ("Fix copy constructors for several ChLink classes") on 2024-07-15, three
days after you opened this, and nobody connected it back to your report. Sorry about that. The current
constructor sets `m_jacobians = nullptr` unconditionally and assigns `m_variables` on both branches.

Confirmed by rebuilding with just that constructor reverted: your snippet crashes at map construction,
and with current `main` it does not.

### The serialization failure is real and is not fixed

It no longer crashes, which is arguably worse, because it now fails silently. Serializing
`std::map<std::string, ChLinkTSDA>` today writes a 254 byte record containing only the `ChObj` base
fields. Every spring parameter and both bodies are missing, and there is no `_version_ChLinkTSDA` at
all. You get a valid-looking JSON file describing nothing useful.

[...]

PR #774, opened 2026-07-27
Fix ChArchiveOut serializing a dangling temporary for std::map and st…

### LINKED ISSUE #776: `ChArchiveOut` serializes a dangling temporary for `std::map` and `std::unordered_map`
Serializing a `std::map` or `std::unordered_map` whose value type is a polymorphic Chrono object
silently writes a truncated record. Reported originally as a crash in #507. A fix is proposed in #774.

### Mechanism

`ChNameValue` is non-owning. Its only value constructor takes `const T&` and stores a raw pointer,
casting away constness on the way ([ChArchive.h#L301-L308](https://github.com/projectchrono/chrono/blob/main/src/chrono/serialization/ChArchive.h#L301-L308)):

```cpp
ChNameValue(const std::string& mname, const T& mvalue, ...)
    : _name(mname), _value((T*)(&mvalue)), ...
```

Because the parameter is `const T&`, a temporary binds to it happily, and the stored pointer dangles
as soon as that temporary dies.

Dereferencing a `std::map<T, Tv>` iterator yields `std::pair<const T, Tv>&`. But the `out()` wrappers
for both associative containers declared their `ChNameValue` with a **non-const** key
([ChArchive.h#L995](https://github.com/projectchrono/chrono/blob/main/src/chrono/serialization/ChArchive.h#L995)
and [#L1011](https://github.com/projectchrono/chrono/blob/main/src/chrono/serialization/ChArchive.h#L1011)):

```cpp
ChNameValue<std::pair<T, Tv>> array_key(std::to_string(i), (*it));   // <-- non-const key
this->out(array_key);
```

The two types differ, so `std::pair`'s converting constructor materializes a temporary
`std::pair<T, Tv>`. Two consequences follow:

1. The archive writes a **copy** of the element rather than the element itself.
2. That copy is destroyed at the closing semicolon, so the next line reads freed memory.

`std::vector` and `std::list` are unaffected: they pass `bVal.value()[i]` and `(*iter)`, which are
already exactly `T&`, so no conversion and no temporary occur.

### Observed effect

Serializing `std::map<std::string, ChLinkTSDA>` to JSON produces a 254 byte record containing only
the `ChObj` fields. There is no `_version_ChLinkTSDA`, no `m_k`, `m_r`, `m_f`, `m_rest_length`, and no
bodies. The correct record is about 1100 bytes. Measured container matrix, with each value stamped
`rest_length = 12345` and `k = 999`:

| Container | Result |
|---|---|
| `vector<ChLinkTSDA>` | full record, 1006 B, sentinels present |
| `list<ChLinkTSDA>` | full record, 1006 B, sentinels present |
| `pair<string, ChLinkTSDA>` used directly | full record, 1028 B, sentinels present |
| `map<string, ChLinkTSDA>` | **truncated to `ChObj`**, 254 B, sentinels lost |
| `unordered_map<string, ChLinkTSDA>` | **truncated to `ChObj`**, 254 B, sentinels lost |
| `map<string, ChVector3d>` | full and correct |

### Why this went unnoticed for so long

Reading a destroyed object is undefined behavior, and undefined does not mean reliably broken. A
trivially destructible type such as `ChVector3d` normally still reads back its old bytes and looks
perfectly correct, which is why the last row above passes. Only a type with a non-trivial destructor,
and especially a polymorphic one, loses its derived data.

Coverage reinforces this. `utest_CH_archive` had no STL container test of any kind, and the only map
in `demo_CH_archive` is an `unordered_map<int, double>` of PODs, serialized in a function whose
deserialization half contains no assertions at all.

### Age

The `unordered_map` wrapper has had this shape since ef4ef487f5 (2017-02-09, "Archive machinery for
(de)serializing std::unordered_map"). Notably the `in()` half of that same commit was written
correctly, with a named local `std::pair<T, Tv> mpair;`, and it is still correct today. The
out-versus-in asymmetry has been there from the beginning, which suggests an oversight rather than a
deliberate choice. The `std::map` wrapper was added later and copied the pattern.

### On fixing it

Worth recording, because the obvious fix is wrong: giving the temporary a name so that it outlives the
`out()` call does compile and does stop the crash, but it keeps serializing a **copy**. That is not
equivalent, because `ChLink`'s copy constructor nulls both body pointers (see #777), so it would
trade a loud bug for a silent one.

#774 instead declares the key `const` in the two `out()` wrappers so the `ChNameValue` binds to the
real element and no temporary is created. The `in()` path is deliberately untouched:
`_wrap_pair::ArchiveIn` writes into `first`, so a const key there would assign through a pointer to a
const object. Since template member functions are instantiated only when used, `ArchiveIn` is never
instantiated on the const-keyed specialization.

### Related

A separate instance of the same non-owning-pointer mistake in this subsystem is filed as #779.

DanNegrut:
Fixes #776

…d::unordered_map

Dereferencing a std::map<T, Tv> iterator yields std::pair<const T, Tv>&, but the out() wrappers for std::map and std::unordered_map declared their ChNameValue with a non-const key. The two types differ, so the compiler materialized a temporary pair, and ChNameValue stores only a raw pointer to its value (it takes const T& and keeps (T*)&mvalue). Two things followed: the archive wrote a copy of the element rather than the element itself, and that copy had already been destroyed by the time it was written, so the write read freed memory.

The visible effect depended on the value type. A trivially destructible type such as ChVector3d usually still read back its old bytes and looked correct, which is why this survived so long. A polymorphic type with a non-trivial destructor lost everything below its base class: serializing a std::map<std::string, ChLinkTSDA> produced a 254 byte record holding only the ChObj fields, with no _version_ChLinkTSDA, no spring coefficients and no bodies. Reported in #507, where it originally presented as a segmentation fault.

The fix declares the key const in the two out() wrappers, so the ChNameValue binds directly to the real map element and no temporary is created. The in() path is deliberately left alone: _wrap_pair::ArchiveIn writes into first, so a const key there would assign through a pointer to a const object. Because member functions of a class template are instantiated only when used, ArchiveIn is never instantiated on the const-keyed specialization, and only ArchiveOut, which merely reads first, is affected.

The naive alternative, giving the temporary a name so that it outlives the call, would compile and stop the crash but would keep serializing a copy. That is not equivalent: ChLink's copy constructor nulls both body pointers, so it would trade a loud bug for a silent one.

Adds five regression tests to utest_CH_archive. They assert that the archive does not copy the container element, which is deterministic, rather than asserting on how the resulting undefined behavior happens to manifest, which varies with type and optimizer. vector and list are included as controls. With the two-line change reverted, the map and unordered_map tests fail (one copy observed, and the value reads back as the test payload's destructor poison) while the vector and list controls still pass. This file previously had no STL container coverage at all, and the only map in demo_CH_archive is an unordered_map<int, double> of PODs with no assertions, which is why nothing in tree caught this.

Unrelated and pre-existing, noted here only because this fix makes it visible in serialized output: ChObj's copy constructor never assigns m_tag, so any copied or cloned Chrono object carries a garbage tag. Confirmed without any serialization involved, where a ChBody tagged 77 reports 32765 after a copy. That is left for a separate report.

Verified with clang (ROCm 7.1) on Windows, Release: all 10 core unit test binaries pass (44 cases) and demo_CH_archive still round-trips.


rserban:
@DanNegrut - I assigned Dario Mangoni as a reviewer. The archiving stuff was implemented at Parma and I am not very familiar with it.

dariomangoni:
The proposed unit test was over complicated and also was set to avoid hitting the other issue https://github.com/projectchrono/chrono/pull/775

The change to `const T` was indeed correct and merged.

PR #775, opened 2026-07-27
Make the JSON and XML archives round trip floating point values exactly

### LINKED ISSUE #778: JSON and XML archives lose precision on every `double` (6 significant digits)
`ChArchiveOutJSON` and `ChArchiveOutXML` write floating point values at the default `std::ostream`
precision, which is **6 significant digits**. A `double` needs 17 to round trip bit-exactly. So saving
a model to JSON or XML and loading it back does not reproduce the model. Nothing warns.

The write is a plain insertion with no precision set, for example
[ChArchiveJSON.cpp#L64](https://github.com/projectchrono/chrono/blob/main/src/chrono/serialization/ChArchiveJSON.cpp#L64):

```cpp
void ChArchiveOutJSON::out(ChNameValue<double> bVal) {
    ...
    m_ostream << bVal.value();
}
```

Neither `ChArchiveJSON.cpp` nor `ChArchiveXML.cpp` contains any `precision`, `setprecision`, or
`max_digits10`.

### Measured

The binary backend serves as a built-in control, since it writes the raw bytes rather than printing
them:

| Original | JSON | XML | Binary |
|---|---|---|---|
| `0.33333333333333331` | `0.333333` **lossy** | `0.333333` **lossy** | exact |
| `12345.678901234567` | `12345.7` **lossy** | `12345.7` **lossy** | exact |

Note the magnitude dependence in the second row. The absolute error there is about 0.022, because six
significant digits is a *relative* budget. Larger coordinates lose proportionally more.

### Why the test suite does not catch it

`utest_CH_archive`'s round-trip tests compare with `ASSERT_NEAR` at `ABS_ERR = 1e-5` on states of
magnitude around 1, where 6 significant digits is coincidentally good enough. The tolerance was
presumably chosen for integrator drift, and it happens to mask this.

### Fix

A fix is proposed in #775. It sets the stream precision to `max_digits10` in each output
archive's constructor and restores the caller's setting in the destructor, since the stream belongs to
the caller and `std::cout` should not be left permanently reconfigured.

The float overloads use 9 rather than inheriting the double's 17. Printing a float with eight digits
more than it carries also round trips, but it exposes binary noise: `0.6f` would be written as
`0.60000002384185791` instead of `0.600000024`.

`ChArchiveOutASCII` is left alone. It has no reader, so it cannot round trip, and it is a human-facing
dump where 6 digits is a reasonable default.

### The readability cost, which is the real decision here

Stating this plainly, because it is presumably why the precision was never raised. Any value that is
not exactly representable in binary now prints in full:

| Value | At 17 significant digits |
|---|---|
| `0.5` | `0.5` |
| `2.0` | `2` |
| `0.1` | `0.10000000000000001` |
| `9.81` | `9.8100000000000005` |
| `0.05` | `0.050000000000000003` |

On the demo's system archive this changed the file size by under one percent, so size is not really
the issue; legibility is.

There is an option that is both exact and tidy: emit the shortest decimal string that round trips,
which renders `0.1` as `0.1`. `std::to_chars` does exactly that, but its floating point overloads are
not portable enough across the compilers Chrono supports, particularly older libstdc++ and libc++, so
I did not use it. A portable equivalent is to try increasing precision until a parse-back matches,
which costs a small number of format-and-parse cycles per value. I would be happy to redo the patch
that way if you would rather keep the files legible; I went with the simple version first because it
is obviously correct and easy to review.

DanNegrut:
Fixes #778

ChArchiveOutJSON and ChArchiveOutXML wrote floating point values at the default ostream precision, which is 6 significant digits, while a double needs 17 to be recovered exactly. Saving a model to JSON or XML and loading it back therefore returned different numbers, silently. Measured, with the binary backend as a control since it writes raw bytes rather than printing them:

  original                JSON        XML         binary
  0.33333333333333331     0.333333    0.333333    exact
  12345.678901234567      12345.7     12345.7     exact

Note the second row: six significant digits is a relative budget, so the absolute error grows with magnitude, reaching about 0.022 there. Anyone using JSON or XML to checkpoint and restart a simulation was not restarting from where they stopped.

The fix raises the precision to max_digits10, which is the number of decimal digits that uniquely distinguishes every value of a type. It is set on the stream in the output archive's constructor and restored in the destructor, because the stream belongs to the caller and may be reused afterwards; std::cout, for one, should not be left permanently reconfigured. There is a test for that restoration.

The float overloads set 9 rather than inheriting the double's 17. Printing a float with eight digits more than it carries also round trips, but it exposes binary noise: 0.6f would be written as 0.60000002384185791 instead of 0.600000024. There is a test guarding against that regression too, since exactness alone does not catch it.

Readability cost, stated plainly because it is the reason this was not simply always correct: any value that is not exactly representable in binary now prints in full. 0.1 becomes 0.10000000000000001 and 9.81 becomes 9.8100000000000005, while 0.5 and 2.0 are unaffected. On the demo's system archive this changed the file size by less than one percent. The alternative that is both exact and tidy is to emit the shortest representation that round trips, which would render 0.1 as 0.1. std::to_chars does that directly, but its floating point overloads are not portable enough across the compilers Chrono supports, particularly older libstdc++ and libc++, so it is left as a possible follow-up rather than done here.

ChArchiveOutASCII is deliberately untouched. It has no reader, so it cannot round trip and is a human-facing dump where 6 digits is a reasonable default.

New test file utest_CH_archive_precision, in its own translation unit so that it does not collide with the additions PR #774 makes to utest_CH_archive. With the four source changes reverted, the four JSON and XML cases fail and the binary control still passes. All 11 core unit test binaries pass, including the pre-existing JSON, XML and Binary Fourbar round trips, which confirms the readers accept the longer values. demo_CH_archive still round-trips.

### DIFF
--- a/src/chrono/serialization/ChArchiveJSON.cpp
+++ b/src/chrono/serialization/ChArchiveJSON.cpp
@@ -1,8 +1,18 @@
 #include "chrono/serialization/ChArchiveJSON.h"
 
+#include <limits>
+
 namespace chrono {
 
 ChArchiveOutJSON::ChArchiveOutJSON(std::ostream& stream_out) : m_ostream(stream_out) {
+    // Write floating point values with enough digits to be read back bit-exactly. The default
+    // ostream precision is 6 significant digits, which silently discards about ten digits of a
+    // double and makes a save/load cycle lossy. max_digits10 is the number of decimal digits that
+    // uniquely distinguishes every value of the type (17 for double).
+    // The caller owns the stream and may well be reusing it (std::cout, for instance), so the
+    // previous precision is recorded here and put back by the destructor.
+    m_precision_saved = m_ostream.precision(std::numeric_limits<double>::max_digits10);
+
     m_ostream << "{ ";
     ++tablevel;
 
@@ -17,6 +27,8 @@ ChArchiveOutJSON::~ChArchiveOutJSON() {
     is_array.pop();
 
     m_ostream << "\n}" << std::endl;
+
+    m_ostream.precision(m_precision_saved);
 }
 
 void ChArchiveOutJSON::indent() {
@@ -57,7 +69,12 @@ void ChArchiveOutJSON::out(ChNameValue<float> bVal) {
     if (is_array.top() == false)
         m_ostream << "\"" << bVal.name() << "\""
                   << "\t: ";
+    // The stream is set to double precision for the archive's lifetime, which would print a float
+    // with eight digits more than it actually carries, exposing binary noise: 0.6f would be written
+    // as 0.60000002384185791. Nine digits is what uniquely identifies a float.
+    const std::streamsize saved = m_ostream.precision(std::numeric_limits<float>::max_digits10);
     m_ostream << bVal.value();
+    m_ostream.precision(saved);
     ++nitems.top();
 }
 
--- a/src/chrono/serialization/ChArchiveJSON.cpp
+++ b/src/chrono/serialization/ChArchiveJSON.cpp
@@ -1,18 +1,11 @@
 #include "chrono/serialization/ChArchiveJSON.h"
 
 #include <limits>
+#include <charconv>
 
 namespace chrono {
 
-ChArchiveOutJSON::ChArchiveOutJSON(std::ostream& stream_out) : m_ostream(stream_out) {
-    // Write floating point values with enough digits to be read back bit-exactly. The default
-    // ostream precision is 6 significant digits, which silently discards about ten digits of a
-    // double and makes a save/load cycle lossy. max_digits10 is the number of decimal digits that
-    // uniquely distinguishes every value of the type (17 for double).
-    // The caller owns the stream and may well be reusing it (std::cout, for instance), so the
-    // previous precision is recorded here and put back by the destructor.
-    m_precision_saved = m_ostream.precision(std::numeric_limits<double>::max_digits10);
-
+ChArchiveOutJSON::ChArchiveOutJSON(std::ostream& stream_out, bool full_precision) : m_ostream(stream_out), m_full_precision(full_precision) {
     m_ostream << "{ ";
     ++tablevel;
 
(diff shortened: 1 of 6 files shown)


fallenmi (changes requested):
Disclosure: I used Codex to help prepare this review; I independently built and ran the tests described below against exact head `78c056f031b4cff88a37e5b0113661fca9a678e7` and the current merge.

Requesting changes because the JSON path still does not provide the bit-exact round trip this PR promises. The new test already includes `1.2345678901234567e+8`, but `EXPECT_DOUBLE_EQ` permits a small ULP difference. With that value, exact head/current merge serialize `123456789.01234567`, then `ChArchiveInJSON` returns bits `0x419d6f34540ca459` instead of the source bits `0x419d6f34540ca458` (one ULP away), so the added test passes while the value is not exact.

`ChArchiveInJSON` still calls RapidJSON `document.Parse<0>`. A focused parser control using the vendored RapidJSON reproduced the wrong `...459` result with `Parse<0>` and returned the original `...458` bits with `kParseFullPrecisionFlag`. Please make the JSON reader parse at full precision (or otherwise guarantee the exact result) and change the regression assertion to compare the bit pattern rather than the ULP-tolerant `EXPECT_DOUBLE_EQ`.

For context, the six added tests pass on both head and current merge, while a 14-value magnitude/exponent oracle improves JSON from 8/28 exact scalar checks on base to 25/28 on head/merge but still catches this ordinary finite-value mismatch.

dariomangoni:
@fallenmi since the `kParseFullPrecisionFlag` is expected to run much slower than the default flags I'm more prone to add a templatization to just the constructor of `ChArchiveInJSON `instead of hard-coding the flag.

However, if ULP precision is requested, shouldn't we update also the XML archiver to do std::from_chars instead of std::stod?

dariomangoni:
Here is the final form of the ChArchive structures. I let the user pick full precision (default) but also to decide for a lower precision (for those cases in which the full precision would have led to performance or size degradation).

In case of full precision the following functions are used
|   | Writing | Reading |
| ------------- | ------- | ------ |
| JSON  | `to_chars` | `GetDouble` |
| XML  | `to_chars`| `from_chars` |

If needed I can comment more on the specific reason for picking these functions.

The usage of `to_chars` leads to write exactly the number of significant digits required to achieve the desired accuracy.
e.g. 0.6 will be kept 0.6 not 0.60000000000000000

When the full precision is not required the ChArchive class will let the user decide the precision by using stream.setprecision on its side. Surely, we might lose, for this lower-precision case, the distinction between float and double, but I don't consider it as dramatic.

Unit testing is now checking for bit correctness (with full precision enabled).

Issue #781, opened 2026-07-28
FSI-SPH: parameter changes after Initialize() never reach the device (silent no-op setters; broken CRMTerrain moving patch); rheology-failure check is dead code


DanNegrut:
Auditing `src/chrono_fsi/sph` turned up one root cause with two user-visible consequences,
plus an unrelated dead error path. All three have fixes with tests, linked below.
Everything here was verified against `main` at `05de72f4c`, built with gcc 12.2 / ROCm 7.2
on an MI350X (gfx950), Release, in both `CH_USE_SPH_DOUBLE=OFF` and `ON`.

## Root cause

Device kernels read a per-translation-unit `__constant__ ChFsiParamsSPH paramsD`. It is
uploaded only during `Initialize()` (`SphFluidDynamics.cu`, `SphCollisionSystem.cu`,
`SphForceWCSPH.cu`, `SphForceISPH.cu`, `SphBceManager.cu`), and no code path re-uploads it
afterwards. Anything that writes `m_paramsH` after `Initialize()` therefore has no effect
on the simulation.

### Consequence 1: configuration setters are silent no-ops after Initialize()

About thirty public setters on `ChFsiFluidSystemSPH` write only `m_paramsH`. Called after
`Initialize()`, they are accepted and do nothing, with no warning.

Demonstration (periodic fluid box, zero gravity): request
`SetGravitationalAcceleration(0, 0, -9.81)` after `Initialize()` and step 200 times, then
compare against a run where nothing was requested. Particle positions are identical to the
last bit (max |dpos| = 0 m). The same value set *before* `Initialize()` produces free fall
(max |v| = 0.392401 m/s, against g*t = 0.3924).

Fix: #782.

### Consequence 2: the CRMTerrain moving patch does not work

`CRMTerrain::Synchronize` shifts the computational domain mid-run through
`SetComputationalDomain` (`CRMTerrain.cpp:166`). That shift never reaches the device, so
neighbor-search binning keeps using the initial domain and `calcHashD` aborts the run once
the patch travels roughly 40 kernel lengths beyond it.

Demonstration (headless port of `demo_VEH_CRMTerrain_MovingPatch`: same terrain, same
tracked body at 5 km/h, just driven farther): the run dies at about 1.5 m of travel with

```
Error flag intercepted in .../SphCollisionSystem.cu:344 from calcHashD
```

The bundled demo does not surface this because its body never travels that far. With the
fix, the same test completes 4.17 m with 34 patch relocations.

[...]

PR #784, opened 2026-07-28
FSI-SPH: make the rheology-failure error flag actually work

DanNegrut:
Part of #781.

`SphFluidDynamics.cu` declares a host stack `bool error_occurred`, passes it **by
reference** into the `__global__` kernels `EulerStep_D` / `MidpointStep_D` and on into the
device function `TauEulerStep`, then reads it back on the host and calls
`gpuThrowError("Rheology model failed")`. Device code cannot write a host stack variable
and never assigns it, so the check is unreachable. Within the `m_check_errors` block,
positions and densities have working `thrust::any_of` NaN scans, but the updated CRM stress
state has none, and that is the state that went non-finite in the MCC overflow fixed by
#766.

## Change

- Replace the host reference with a device-resident flag, using the idiom this module
  already uses for `calcHashD` (`gpuMallocErrorFlag` / `gpuResetErrorFlag` /
  `gpuCheckErrorFlag` in `SphUtilsDevice.cuh`): the kernels take `volatile bool*`,
  `TauEulerStep` raises it when the updated stress state is not finite, and the host reads
  it back and throws with the kernel named.
- Allocation, reset and readback happen only when `m_check_errors` is enabled; otherwise
  the kernel receives `nullptr` and the entire predicate is skipped on device, so error
  checking costs nothing when it is turned off.
- The predicate covers `tau` and `rho/p` under both rheology models, and the consolidation
  state `pcEvSv` only under MCC. mu(I) neither reads nor updates `pcEvSv` (every access to
  it inside `TauEulerStep` is in the MCC branch), and that state is routinely non-finite
  straight out of default initialization, because `AddSphParticle` derives the specific
  volume with `log(pc / p1)` for the default `pc = 0` (item 1 of #781). Checking it under
  mu(I) aborts healthy runs.
- Same functions, related: `MidpointStep`'s NaN scans ran over the whole array while
  `EulerStep`'s stop at `numActive`, which is what the kernels actually touch. Unified on
  `numActive`.
- New unit test `utest_FSI-SPH_rheology_error_flag`: a CRM bed created with the default
  particle pressure starts with a non-finite consolidation state, and none of mu(I), MCC,
  or MCC with a depth-based initial pressure may be aborted by it. The mu(I) case is a
  regression guard for the false abort described above.

## Scope of the consolidation-state term, stated precisely

Under MCC, both consolidation components that the branch updates are clamped at the end of
it: `pcEvSv.x = fmax(Real(100.0), pcEvSv.x)` and `pcEvSv.z = fmax(Real(1.0), pcEvSv.z)`.
Since `fmax` returns the non-NaN operand, a NaN in either component is replaced by the
clamp floor before the predicate runs. In practice, therefore, the `pcEvSv` term can only
catch a non-finite volumetric strain rate (`pcEvSv.y`, which is not clamped). We mention it
because the clamping also means a nonsense initial specific volume is silently replaced by
1.0 rather than reported, which may be worth a separate look.

## Does the flag fire?

The plumbing is proven end to end by a forced-fire build (the flag raised unconditionally in
`TauEulerStep`), which aborts on the first step with
`Error flag intercepted in .../SphFluidDynamics.cu from TauEulerStep (rheology model
failure)`. The predicate is also demonstrably live on real data: with the earlier, broader
version of it, a CRMTerrain traverse built through `ChFsiProblemSPH` was aborted on step 1
by exactly this path, which is how the initialization issue in item 1 of #781 was found.

Being straightforward about the gap: there is no test that drives the shipped predicate to
fire from the public API. Any input that makes the stress state non-finite also makes
positions non-finite within the same step, and the host-side position scan reports first;
and the obvious deterministic candidate, a poisoned consolidation state under MCC, is
removed by the clamps described above before the predicate sees it.

## Validation

gcc 12.2 / ROCm 7.2, MI350X (gfx950), Release, both precisions. The new unit test passes in
both. A geostatic settling benchmark (106k particles) is unchanged to every printed digit;
wall clock 81-83 s against a 78 s baseline, inside the 78 to 83 s run-to-run spread measured
over eight runs.

Not tested: the CUDA backend (the `gpu*` macros are this module's own backend-portability
layer, used identically by `calcHashD` on both backends), and the ISPH path (these are the
explicit WCSPH integrators).

Deliberately left for follow-ups, since both are pre-existing and module-wide rather than
specific to this change: the per-step flag allocation could become a persistent member (it
also leaks on the throwing paths, which are aborting anyway), and the `SphUtilsDevice.cuh`
error-flag macros ignore their `[[nodiscard]] hipError_t` returns at every call site. Glad
to do either in a separate PR if you want them.

### DIFF
--- a/src/chrono_fsi/sph/physics/SphFluidDynamics.cu
+++ b/src/chrono_fsi/sph/physics/SphFluidDynamics.cu
@@ -41,7 +41,7 @@ void CopyParametersToDevice_SphFluidDynamics(std::shared_ptr<ChFsiParamsSPH> par
 }
 
 SphFluidDynamics::SphFluidDynamics(FsiDataManager& data_mgr, SphBceManager& bce_mgr, bool verbose, bool check_errors)
-    : m_data_mgr(data_mgr), m_verbose(verbose), m_check_errors(check_errors) {
+    : m_data_mgr(data_mgr), m_verbose(verbose), m_check_errors(check_errors), m_errflagD(nullptr) {
     collisionSystem = chrono_types::make_shared<SphCollisionSystem>(data_mgr);
 
     if (m_data_mgr.paramsH->integration_scheme == IntegrationScheme::IMPLICIT_SPH)
@@ -50,10 +50,12 @@ SphFluidDynamics::SphFluidDynamics(FsiDataManager& data_mgr, SphBceManager& bce_
         forceSystem = chrono_types::make_shared<SphForceWCSPH>(data_mgr, bce_mgr, verbose, m_check_errors);
 
     gpuStreamCreate(&m_copy_stream);
+    gpuMallocErrorFlag(m_errflagD);
 }
 
 SphFluidDynamics::~SphFluidDynamics() {
     gpuStreamDestroy(m_copy_stream);
+    gpuFreeErrorFlag(m_errflagD);
 }
 
 // -----------------------------------------------------------------------------
@@ -338,7 +340,7 @@ __device__ void TauEulerStep(Real dT,
                              Real3& tau_offdiag,
                              Real4& rho_p,
                              Real3& pcEvSv,
-                             bool& error_occurred) {
+                             volatile bool* error_flag) {
     if (paramsD.rheology_model_crm == RheologyCRM::MU_OF_I) {
         Real3 new_tau_diag = tau_diag + dT * deriv_tau_diag;
         Real3 new_tau_offdiag = tau_offdiag + dT * deriv_tau_offdiag;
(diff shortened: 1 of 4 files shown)


DanNegrut:
Rebased onto `main` at 6b1a340b4 (it now contains #782, #783 and #785), since the branch as
filed sat on 05de72f4c and had gone into conflict. Force-pushed with `--force-with-lease`;
the branch is now `f3ecab8da`.

The change itself is unchanged by the rebase. The patch this commit introduces is 370 lines
before and after; the only differences are the blob index line, the hunk-header offsets (all
shifted by exactly 3, the three lines #785 deleted above the first hunk) and one context line
(the `#endif` #785 removed). No added or removed line differs.

Since the rebased combination had not been compiled anywhere, it was re-validated before the
push. AMD MI350X (gfx950), ROCm 7.2, gcc 12.2, Release, both precisions
(`CH_USE_SPH_DOUBLE=OFF` and `ON`):

| check | result |
|---|---|
| `Chrono_fsisph` build, both precisions, plus `Chrono_vehicle` | rc=0 |
| `utest_FSI-SPH_rheology_error_flag` (this PR) | passes, fp32 and fp64 |
| `utest_FSI-SPH_setter_guard` (from #782) | 37 checks, 0 failures, both precisions |
| `utest_FSI-SPH_domain_update` (from #783) | 11 checks, 0 failures, both precisions |
| geostatic settling, 106k particles, fp32 | fit slope / rho*g = 0.991, rms residual 3.6%, 83 s |

The two follow-ups already noted in the description still stand and are deliberately not in
this PR: the shared error-flag macros ignore their allocator return codes (one fix at the macro
definitions covers every call site in the module), and the MCC clamps discussed in #781 mean a
default-pressure Cam-Clay bed still cannot make this flag fire.

PR #798, opened 2026-08-04
FSI-SPH: initialize SPHParameters::shifting_method, fix doc defaults

### LINKED ISSUE #797: FSI-SPH: SPHParameters::shifting_method is never initialized, so the shifting algorithm is whatever is in memory
## 1. `shifting_method` is never initialized

`SPHParameters::SPHParameters()` (`src/chrono_fsi/sph/ChFsiFluidSystemSPH.cpp:520`) initializes 27
members and skips `shifting_method`. There is no default member initializer for it either
(`ChFsiFluidSystemSPH.h:87`), though the comment there says "default: XSPH". So the documented
default is never applied, and a default-constructed `SPHParameters` carries an indeterminate
shifting selection into `SetSPHParameters()`.

I did not find this by reading. Two builds of the same host source, differing only in
`CH_USE_SPH_DOUBLE`, took different branches on that member: a test that left it unassigned
printed the "Shifting method is NONE, but shifting parameters have been modified" warning once in
the single-precision build and not at all in the double-precision build. Every literal compared in
that branch (0.5, 1.0, 3.0) is exactly representable as a float, so precision cannot explain it;
only the branch taken can differ. `ShiftingMethod::NONE` is the zero enumerator, so one build's
stack slot happened to be zero. As a control, every test that does assign the member warned the
same number of times in both builds.

What happens downstream: if the value reads as `NONE`, `SphForceWCSPH.cu:572` skips shifting. If it
is a valid non-`NONE` enumerator you get that algorithm. If it is out of range, it passes the
`!= NONE` test at `:572`, matches no case in the `switch` at `:1987` (which has no `default:`), and
no shifting is applied. So it does not crash, but the algorithm is picked at random and there is no
diagnostic.

Five demos and one unit test in `main` construct `SPHParameters` without assigning it:

```
src/demos/fsi/sph/demo_FSI-SPH_CouetteFlow.cpp
src/demos/robot/viper/demo_ROBOT_Viper_CRM.cpp
src/demos/sensor/demo_SEN_CRM_Rendering.cpp
src/demos/vehicle/terrain/demo_VEH_CRMTerrain_MovingPatch.cpp
src/demos/vehicle/terrain/demo_VEH_CRMTerrain_WheeledVehicle.cpp
src/tests/unit_tests/fsi/sph/utest_FSI-SPH_rheology_error_flag.cpp
```

Fix: add `shifting_method(ShiftingMethod::XSPH)` to the initializer list, which is both the
documented default and what `ChFsiFluidSystemSPH`'s own constructor already assigns at
`ChFsiFluidSystemSPH.cpp:99`.

One question for you rather than a claim: every CRM demo that sets `shifting_method` explicitly
picks `PPST_XSPH`, not `XSPH`. If `XSPH` is not the default you want for CRM, say so and I will use
whatever you prefer. I went with the documented value.

## 2. The `NONE` branch checks one parameter against the wrong default

Separate, one line, same function. `ChFsiFluidSystemSPH.cpp:439`:

```cpp
if (m_paramsH->shifting_xsph_eps != 0.5 || m_paramsH->shifting_ppst_push != 1.0 || m_paramsH->shifting_ppst_pull != 1.0) {
```

`shifting_ppst_push` defaults to **3.0**, in both constructors (`:526` and `:112`) and in the header
comment (`ChFsiFluidSystemSPH.h:93`). So the middle term is always true and anyone selecting `NONE`
is told they modified shifting parameters when they did not. The XSPH branch just below, at `:445`,
compares the same member against `3.0` correctly.

Chrono's own test suite prints this false warning 17 times per run, and two demos
(`demo_FSI-SPH_Compressibility`, `demo_FSI-SPH_Poiseuille_flow`) print it too. Fix is `1.0` ->
`3.0`.

## 3. Two more doc comments disagree with the constructor

Found by comparing every "(default: X)" comment in `SPHParameters` against the initializer list.
Both constructors agree with each other, so it is the comments that are stale. These are also what
the published API reference shows, since it is generated from these comments:

| member | comment says | both constructors set |
|---|---|---|
| `num_proximity_search_steps` | 4 | 1 |
| `shifting_diffusion_A` | 2.0 | 1.0 |
| `kernel_type` | `CUBIC_CPLINE` | `CUBIC_SPLINE` (typo in the comment) |

I am correcting the comments to match the code, since the code is what runs. But
`num_proximity_search_steps` is worth a look: at 1, neighbor lists are rebuilt every step. If 4 was
the intended default, that is a real performance difference and the code should change instead.

Found while auditing `src/chrono_fsi/sph`. A PR for all of this follows.

DanNegrut:
`SPHParameters::SPHParameters()` set 27 of its 28 members and skipped `shifting_method`, which has
no default member initializer either. So a default-constructed `SPHParameters` carried an
indeterminate shifting selection into `SetSPHParameters()`, and the documented default of XSPH was
never applied by anything. One line fixes it.

These construct `SPHParameters` without assigning it, so they were all reading it uninitialized:

```
src/demos/fsi/sph/demo_FSI-SPH_CouetteFlow.cpp
src/demos/robot/viper/demo_ROBOT_Viper_CRM.cpp
src/demos/sensor/demo_SEN_CRM_Rendering.cpp
src/demos/vehicle/terrain/demo_VEH_CRMTerrain_MovingPatch.cpp
src/demos/vehicle/terrain/demo_VEH_CRMTerrain_WheeledVehicle.cpp
src/tests/unit_tests/fsi/sph/utest_FSI-SPH_rheology_error_flag.cpp
```

It turned up because two builds of the same source, differing only in `CH_USE_SPH_DOUBLE`, took
different branches on that member. More detail in the issue.

Also here, since it is the same drift between the doc comments and the constructor:

- The `ShiftingMethod::NONE` branch of `CheckSPHParameters` compared `shifting_ppst_push` against
  `1.0` when the default is `3.0`, so it told everyone who selected `NONE` that they had modified
  shifting parameters. The XSPH branch just below already compared against `3.0`. The unit
  test suite printed that false warning 17 times per run; it is now 0.
- Three comments corrected to match the constructor: `num_proximity_search_steps` said 4 and is 1,
  `shifting_diffusion_A` said 2.0 and is 1.0, `kernel_type` said `CUBIC_CPLINE`.

The comments were changed to match the code rather than the reverse, because the code is what runs
and changing those two numbers would alter behavior for existing users. If either was meant to be
the documented value, that is a separate and deliberate change. Worth knowing for
`num_proximity_search_steps`: at 1, neighbor lists are rebuilt every step.

One thing to check on my side: every CRM demo that sets `shifting_method` explicitly picks
`PPST_XSPH`, not `XSPH`. I used the documented value. If XSPH is not the default you want, say so.

`utest_FSI-SPH_setter_guard` now checks that a default-constructed `SPHParameters` holds these four
documented defaults.

Validated on an MI350X in both single and double precision: all five FSI-SPH unit tests pass in
each, and the false-warning count across the whole suite goes from 17 per run to 0 in both.

On the demos, precisely, since this changes their behavior from "whatever was in memory" to XSPH:
two of the five were exercised, `demo_FSI-SPH_CouetteFlow` and
`demo_VEH_CRMTerrain_WheeledVehicle`, both of which run to completion with no false warning. The
other three were not built in a headless configuration: `demo_ROBOT_Viper_CRM` and
`demo_VEH_CRMTerrain_MovingPatch` are gated on `CH_ENABLE_MODULE_VSG`, and `demo_SEN_CRM_Rendering`
on the sensor module. If your CI covers those, that is the gap worth checking.

Closes #797.

### DIFF
--- a/src/chrono_fsi/sph/ChFsiFluidSystemSPH.cpp
+++ b/src/chrono_fsi/sph/ChFsiFluidSystemSPH.cpp
@@ -436,7 +436,7 @@ void ChFsiFluidSystemSPH::CheckSPHParameters() {
 
     // Check shifting method and whether defaults have changed
     if (m_paramsH->shifting_method == ShiftingMethod::NONE) {
-        if (m_paramsH->shifting_xsph_eps != 0.5 || m_paramsH->shifting_ppst_push != 1.0 || m_paramsH->shifting_ppst_pull != 1.0) {
+        if (m_paramsH->shifting_xsph_eps != 0.5 || m_paramsH->shifting_ppst_push != 3.0 || m_paramsH->shifting_ppst_pull != 1.0) {
             cerr << "WARNING: Shifting method is NONE, but shifting parameters have been modified. These "
                     "changes will not take effect."
                  << endl;
@@ -522,6 +522,7 @@ ChFsiFluidSystemSPH::SPHParameters::SPHParameters()
       initial_spacing(0.01),
       d0_multiplier(1.2),
       max_velocity(1.0),
+      shifting_method(ShiftingMethod::XSPH),
       shifting_xsph_eps(0.5),
       shifting_ppst_push(3.0),
       shifting_ppst_pull(1.0),
--- a/src/chrono_fsi/sph/ChFsiFluidSystemSPH.h
+++ b/src/chrono_fsi/sph/ChFsiFluidSystemSPH.h
@@ -83,7 +83,7 @@ class CH_FSI_API ChFsiFluidSystemSPH : public ChFsiFluidSystem {
         EosType eos_type;                              ///< equation of state (default: ISOTHERMAL)
         ViscosityMethod viscosity_method;              ///< viscosity treatment (default: ARTIFICIAL_UNILATERAL)
         BoundaryMethod boundary_method;                ///< boundary treatment (default: ADAMI)
-        KernelType kernel_type;                        ///< kernel type (default: CUBIC_CPLINE)
+        KernelType kernel_type;                        ///< kernel type (default: CUBIC_SPLINE)
         ShiftingMethod shifting_method;                ///< shifting method (default: XSPH)
         int num_bce_layers;                            ///< number of BCE layers (boundary and solids, default: 3)
         double initial_spacing;                        ///< initial particle spacing (default: 0.01)
@@ -93,7 +93,7 @@ class CH_FSI_API ChFsiFluidSystemSPH : public ChFsiFluidSystem {
         double shifting_ppst_push;                     ///< PPST pushing coefficient (default: 3.0)
         double shifting_ppst_pull;                     ///< shifting beta coefficient (default: 1.0)
         double shifting_beta_implicit;                 ///< shifting coefficient used in implicit solver (default: 1.0)
-        double shifting_diffusion_A;                   ///< shifting coefficient used in diffusion (default: 2.0, range 1 to 6)
+        double shifting_diffusion_A;                   ///< shifting coefficient used in diffusion (default: 1.0, range 1 to 6)
         double shifting_diffusion_AFSM;                ///< shifting coefficient used in diffusion (default: 3.0)
         double shifting_diffusion_AFST;                ///< shifting coefficient used in diffusion (default: 2.0)
         double min_distance_coefficient;               ///< min inter-particle distance as fraction of kernel radius (default: 0.01)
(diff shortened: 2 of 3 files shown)


rserban:
The unit test added with this PR seems pointless. There is no way to enforce that the documented default values match the actual default values.  In the future, one can easily change a default value, make the corresponding change in this unit test, and still forget to update the documentation. 

Unless I am missing something, I will close this PR and make the necessary documentation fixes in a separate commit.

rserban:
Addressed in commit #accbfc6550

PR #800, opened 2026-08-04
Sensor: shade material-less shapes with Chrono's default material

DanNegrut:
A visual shape with no `ChVisualMaterial` was shaded by the OptiX backend from literals inside
`ChOptixPipeline::GetMaterial`, `Kd = 0.5` grey with `roughness = 1`, instead of
`ChVisualMaterial::Default()`, which is white and smooth and is what `ChVisualShape::GetColor()`
reports for an empty material list. Irrlicht and VSG use the shared default; OptiX was the only
backend that did not. The grey was deliberate when written (2f6e628f6, 2021), because the core default
was then an unusable `Kd = (1.0, 2.0, 2.0)`. Core was fixed a year later (45264a525) and this copy was
never revisited, so the two drifted apart.

## Measured (RTX 5090, CUDA 12.8, OptiX 9.0.0)

- A material-less box rendered **2.0011x darker** under OptiX than under a second ray-tracing backend
  given the same scene in the same process: RMSE 11.68, max 68 code units, 14.41 % of pixels off by
  more than 2. After: RMSE 0.52, max 1, **0.00 %** over threshold. Same factor of two then measured on
  material-less spheres, cylinders and an explicit triangle mesh, each reaching the default through a
  different function.
- Two standard helpers disagreed with each other, no second backend involved:
  `chrono::utils::AddBoxGeometry` attaches `Default()` by default argument and rendered white;
  `ChBodyEasyBox` attaches nothing and rendered grey.
- Strongest check, and it needs no second backend: box, sphere, cylinder and mesh with no material
  produce a **byte-identical** OptiX frame to those same shapes carrying `Default()` explicitly
  (md5 `0040464e8f25f8bfec8d91b4db4fc833`). Any disagreeing struct field would show up here.

## The change

The explicit-material conversion is extracted into `MakeMaterialParameters` and both arms of
`GetMaterial` call it, so the two cannot drift again. The struct is value-initialized: it has no
default member initializers, and the old default branch omitted `anisotropy` and `theta_p`, which
reached the device as stack garbage. Field coverage is mechanical: 34 declared members, 33 assigned,
the exception being the padding.

The first commit lands separately because it changes nothing a correct program can observe.
`ChVisualMaterial`'s constructor left `use_hapke` and seven `hapke_*` members uninitialized, and
`GetMaterial` copies the seven floats to the device for every material without testing `bsdf_type`, so
every render copied indeterminate host memory to the GPU. Latent, since the shader reads them only
under the Hapke BSDF, but undefined, and this change adds a caller. The same commit makes `Default()`'s
lazy init thread-safe by moving the static inside the function: ThreadSanitizer reports **1038 data
races over 30 runs** of an 8-thread first-access probe against the old namespace-scope `if (!ptr)`
pattern and **none over 30 runs** against the new one, and the old pattern returned more than one
distinct "singleton" in all 30 runs, worst case 7 of 8 threads holding different objects.

## Behaviour change, deliberate

Scenes containing shapes with no explicit material now render brighter and glossier, including scenes
deserialized from files saved with an empty material list. `CHANGELOG.md` records this under
Unreleased with the three lines that restore the old look, and the Sensor manual gains a line. Shapes
carrying an explicit material are bit-identical before and after, on six scene configurations.

## Tests

Two CPU-only binaries, outside the OptiX block so they run without a GPU.
`utest_SEN_default_material` (12 tests) pins the default's values, initialization of every member,
serialization round trips including the 17 members `ArchiveOut` does not write, and that colouring a
shape cannot mutate the shared singleton. `utest_SEN_default_material_firstaccess` holds one test
alone in its process, because `Default()` initializes on first call and a concurrent-first-access test
only has a race to observe if nothing else has called it yet: against the old code it fails **200 of
200** runs alone and **0 of 200** sharing a binary. Sensor suite 13 of 13; whole-tree build clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Change Log
 ==========
 
 - [Unreleased (development branch)](#unreleased-development-branch)
+  - [\[Changed\] Chrono::Sensor appearance of shapes with no visual material](#changed-chronosensor-appearance-of-shapes-with-no-visual-material)
 - [Release 10.0.0 (2026-03-27)](#release-1000-2026-03-27)
   - [\[Changed\] Python support in conda packages](#changed-python-support-in-conda-packages)
   - [\[Added\] PyChrono-NumPy integration](#added-pychrono-numpy-integration)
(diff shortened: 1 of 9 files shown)


rserban:
Merge deferred until other branches that affect Chrono::Sensor are dealt with.

PR #803, opened 2026-08-06
Core: link the CUDA runtime, which Chrono_core already calls

DanNegrut:
The multicore collision sources reach the CUDA runtime API through gpu/ChGpuRuntime.h, so libChrono_core carries an undefined reference to cudaGetErrorString and never links cudart. It resolves by accident in builds where some other target pulls the runtime in, and fails where none does:

  /usr/bin/ld: libChrono_core.so: undefined reference to `cudaGetErrorString`

Reproduced with -DCH_ENABLE_MODULE_SENSOR=ON -DCH_USE_SENSOR_OPTIX=OFF on a machine that has CUDA installed, which is what one would configure to exercise a non-OptiX rendering path on an NVIDIA box. That configuration links after this change.

### DIFF
--- a/src/chrono/CMakeLists.txt
+++ b/src/chrono/CMakeLists.txt
@@ -1538,6 +1538,14 @@ if(CHRONO_THRUST_FOUND)
   endif()
 endif()
 
+if(CHRONO_CUDA_FOUND)
+  # The multicore collision sources reach the CUDA runtime API through gpu/ChGpuRuntime.h, so
+  # Chrono_core itself references cudaGetErrorString. Declare that here rather than leaving the
+  # symbol for a consumer to satisfy by accident: it resolves only in builds where something else
+  # already links the runtime, and fails in those where nothing does.
+  target_link_libraries(Chrono_core PRIVATE CUDA::cudart)
+endif()
+
 if(CH_ENABLE_OPENMP)
   target_link_libraries(Chrono_core PUBLIC OpenMP::OpenMP_CXX)
 endif()


rserban:
I cannot reproduce this.

Where/how is `gpu/ChGpuRuntime.h` ever included in the multicore collision detection system?

DanNegrut:
Closing this.

I traced the premise and it does not hold. `chrono/gpu/ChGpuRuntime.h` is never included from core: across the tree it is included by 19 files in `chrono_dem` and 7 in `chrono_fsi`, plus one header-to-header include of `ChGpuPrimitives.h`. None of the 18 files in `src/chrono/collision/multicore/` include anything CUDA related.

Even if one of them did, the CUDA body of that header sits behind `#if defined(__CUDACC__) || defined(CHRONO_USE_CUDA)`. Core compiles no `.cu` sources, so `__CUDACC__` is never set for it, and `CHRONO_USE_CUDA` is defined only as `PUBLIC` on `Chrono_dem` and `Chrono_fsisph`. `Chrono_core` therefore cannot reference `cudaGetErrorString`, and the rationale given in the PR description was wrong.

The link error that prompted this was real, but it cannot have originated in core on main, so a stale build tree is the most likely explanation. If it resurfaces, the useful thing to capture is the full link command line, which will name whichever target actually pulled the symbol in.

One incidental observation from the trace: `CHRONO_USE_CUDA` is `PUBLIC` on both `Chrono_dem` and `Chrono_fsisph`, so any dependent target that compiles a plain `.cpp` including that header inherits the macro and then needs `cudart` itself. Not proposing a change for it here, just noting it.

---

## Backing evidence

Verified by reading committed source on `main` (commit `dab1ccb`), via `git grep` against
the commit rather than a working tree:

1. `src/chrono/gpu/` contains exactly two files, `ChGpuPrimitives.h` and `ChGpuRuntime.h`.
2. Includes of `chrono/gpu/*` across `src/`: 19 from `chrono_dem`, 7 from `chrono_fsi`,
   and 1 from within `src/chrono/` which is `ChGpuRuntime.h` including `ChGpuPrimitives.h`.
   Zero from any core `.cpp`.
3. All 18 files in `src/chrono/collision/multicore/` include nothing CUDA or gpu related.
4. `CHRONO_USE_CUDA` is defined in exactly two places, both `PUBLIC` on other targets:
   `src/chrono_dem/CMakeLists.txt:166` and `src/chrono_fsi/sph/CMakeLists.txt:262`.
5. Core compiles no `.cu`/`.cuh` sources, so `__CUDACC__` is never defined for it.
6. The two mechanisms that could have hidden an include were checked and are clean:
   the core PCH (`ChCorePCH.h`) includes only STL headers plus `chrono/core/ChMatrix.h`,
   and the Thrust block sets `THRUST_DEVICE_SYSTEM_OMP` or `_CPP`, never CUDA.

Caveat: this is a source-level analysis, not a build.

PR #805, opened 2026-08-12
spelling and grammar

jasonnicholson:
Couple of spelling and grammar fixes.


rserban:
Thank you for carefully checking the documentation and pointing out these spelling mistakes. 
However, we perform periodic spellchecks on the code and documentation and prefer to fix those all at once rather than having many small commits or PRs. I made note of the two issues you found and will make sure they are fixed next time we do a spellcheck pass.

jasonnicholson:
10-4

PR #755, opened 2026-06-29
feat(vehicle): optional HIP SCM contact-force backend (MI300X / gfx942)

amd-pratmish:
### Summary

Single PR combining the SCM terrain HIP port (foundation + E2E perf + device body reduction), per maintainer request.

Adds an **opt-in HIP path** for the SCM Bekker / Mohr-Coulomb / Janosi contact-force loop in `SCMLoader::ComputeInternalForces()`. Ray casting and contact-patch BFS stay on CPU.

**Foundation**
- CMake: `CH_ENABLE_VEHICLE_SCM_GPU=ON` in `src/chrono_vehicle/CMakeLists.txt` (OpenCRG pattern) + `CHRONO_SCM_GPU_LIB_DIR`
- In-tree API headers `SCMGpu.h` / `SCMGpuTypes.h`; external HIP impl in `scm_gpu_core` (`FindScmGpu.cmake`)
- Runtime: `CHRONO_SCM_GPU=1` (uniform soil, rigid `ChBody` contactables)
- Auto-fallback to CPU when `hits.size() < CHRONO_SCM_GPU_MIN_HITS` (default **8192**)
- Host compiler stays **g++**; HIP device code via CMake HIP language. Set `-DCMAKE_HIP_ARCHITECTURES=gfx942` (MI300X) or `gfx90a` (MI210).

**E2E perf**
- Pass ray-cast `hits` map directly to `ComputeContactForcesGpu` (no intermediate vector copy)
- Call `scm_gpu::PrimeBuffers()` from all `SCMTerrain::Initialize` overloads
- Dense per-body force accumulation in `SCMTerrainGpu.cpp`

**Device body reduction + async bridge**
- Grid-only host scatter; body forces from device reduction (`scm_reduce_body_forces_kernel`)
- Async HIP streams for pack / compute / scatter (`CHRONO_SCM_GPU_ASYNC`)

### OpenMP → HIP split

```text
[CPU] Update active domains
[CPU] OpenMP ray cast → hits + patch_oob
[CPU] BFS contact patches
[CPU] Pack batch (OpenMP parallel for)
[GPU] scm_compute_forces_kernel  (Bekker + Mohr-Coulomb + Janosi per hit)
[GPU] scm_reduce_body_forces_kernel (v3)
[CPU] Scatter grid / apply ChLoadBodyForce
```

[...]

PR #770, opened 2026-07-21
Document FSI-SPH ROCm/HIP tuning on AMD Instinct GPUs

amd-pratmish:
### Summary

Adds documentation for **ROCm / HIP performance tuning** of Chrono::FSI-SPH demos on AMD Instinct GPUs (MI300X gfx942, MI350X gfx950). No physics or solver source changes — this PR is **documentation only**: build/run instructions, tuning knobs, and **why each change can improve wall-clock time**.

Validated on seven standard FSI-SPH demos: DamBreak, Kernel, ObjectDrop, WaveTank, BaffleFlow, AngleRepose, RassorDrum.

---

### Motivation

FSI-SPH wall time on Instinct is split between **GPU kernels** (forces, neighbor search, activity), **CPU orchestration** (MBS coupling, I/O), and **ROCm runtime overhead** (launches, copies, queues). Default demo settings often leave **particle CSV output** and **every-step neighbor search** enabled, which dominates long fluid runs. This PR documents which knobs remove each bottleneck and how to run tuned examples from the command line.

---

### Files added

| Path | Purpose |
|------|---------|
| `doxygen/documentation/manuals/fsi/fsi_sph_amd_instinct_tuning.md` | Full Doxygen manual (build, run, catalog, per-demo recipes) |
| `doxygen/documentation/manuals/fsi/manual_fsi.md` | Link new page from FSI manual index |
| `docs/FSI_SPH_AMD_TUNING.md` | Plain-Markdown quick reference |
| `docs/fsi_sph_tuning/*.env.example` | Sourceable ROCm / CLI tuning snippets |

---

### Build all FSI-SPH demos

```bash
export ROCM_PATH=${ROCM_PATH:-/opt/rocm}

cmake -S . -B build \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_DEMOS=ON \
  -DENABLE_MODULE_FSI=ON \
  -DCHRONO_HIP=ON \
  -DCMAKE_HIP_ARCHITECTURES=gfx942   # MI300X; use gfx950 for MI350X

cmake --build build --target \
  demo_FSI-SPH_DamBreak demo_FSI-SPH_Kernels demo_FSI-SPH_ObjectDrop \
  demo_FSI-SPH_WaveTank demo_FSI-SPH_BaffleFlow demo_FSI-SPH_AngleRepose \
  demo_FSI-SPH_RassorDrum -j $(nproc)

cd build
[...]

PR #808, opened 2026-08-13
Feat/scm gpu backend (ray casting and contact forces)

Auc7us:
# Move SCM terrain ray-casting and contact forces to the GPU

Branch `feat/scm-gpu-backend`, 20 commits, targets `main`. Previously stacked on
`feat/gpu-backend-hardware-first`, which has since merged as #806; rebased onto current `main`.

**Includes #755** (amd-pratmish), which is still open. That PR's SCM GPU contact-force backend —
`terrain/SCMGpu.h`, `SCMGpuTypes.h`, `SCMTerrainGpu.cpp`, `terrain/gpu/SCMGpuHost.cpp` and
`SCMGpuKernels.hip.cpp` — is carried here byte-identical. The ray-cast backend, the build
integration and everything below are new on top of it. Merging this supersedes #755; it should not
land twice.

## Change

Both per-node-independent stages of an SCM step move to the GPU: ray-casting (one ray per active
grid node against vehicle collision geometry) and contact forces (Bekker / Janosi-Hanamoto at
every hit). Kernels are HIP (`terrain/gpu/*.hip.cpp`) with plain C++ host wrappers, so one source
serves ROCm and, via HIP-on-NVIDIA, CUDA hardware. The CPU implementation is untouched and remains
the fallback.

On by default where hardware and toolchain allow, with no CMake option, environment variable or
API call required:

| | Default | Force off |
|---|---|---|
| Build | `CH_ENABLE_VEHICLE_SCM_GPU=ON` | set `OFF` |
| Contact forces | enabled (`min_hits=8192` CPU-fallback heuristic) | `SetScmGpuEnabled(false)` |
| Ray-casting | enabled | `EnableRaycastGpuHip(false)` |

Ray-casting also needs explicit active domains (`AddActiveDomain`) and triangle-mesh collision
geometry, the only shape type these kernels intersect. Lacking either, the backend declines the
step and the CPU path runs, so the default is safe for models that cannot use it
(`demo_ROBOT_RoboSimian_SCM` is one). Per-step fallback is silent by design;
`GetNumRaycastGpuSteps()` and `GetNumContactForceGpuSteps()` report what actually ran.

Ray-cast kernels are FP32 on every platform so a model behaves the same on AMD and NVIDIA. FP64
stays available through `SCM_RAYCAST_GPU_PRECISION`; the two differ by 3 um of sinkage against the
1.3 cm between the GPU and CPU ray-cast paths. `EnableRaycastGpuReference()` runs the same
algorithm on the CPU as a validation stand-in.

`CHRONO_ENABLE_HIP_ON_NVIDIA` flips to `ON` here (the option itself landed in #806).
These kernels exist only in HIP, so on NVIDIA "is HIP available" and "can this use the GPU" are one
question. DEM and FSI::SPH still resolve to CUDA. Without HIP the feature resolves to `NONE` and
the CPU path runs, reported as a status line, without forcing the option `OFF` in the cache.

Ray-body culling: each body carries a face range and a world AABB that a ray slab-tests before
touching its triangles, and nodes outside every body's (x,y) footprint never become queries at
all. Without it every ray scanned every triangle in the scene, 669M tests per step at 6 obstacles.

## Active-domain crash fix (review this first)

Independent of the GPU work. Fixes a null dereference that crashes `demo_ROBOT_Curiosity_SCM`,
`demo_ROBOT_Curiosity_SCM_Sensor` and `demo_ROBOT_Viper_SCM_Sensor` on current `main` with no GPU
modules enabled.

`SetupInitial()` appends a placeholder domain with `m_body = nullptr`. `AddActiveDomain()` appends
real domains and sets `m_user_domains = true` but never removes the placeholder, so the first step
iterates every domain and dereferences the null body. Calling `AddActiveDomain()` after
`Initialize()` crashes, before it works; that is undocumented and not the order the demos use.

Exposed by `6aba95d7d` (Jul 2026), which added `SetupInitial()` to all three `Initialize()`
overloads. The placeholder dates from `725aa2aca` (Apr 2025) but was dormant until then. The demos
did not change, and the crash only appears after a rebuild.

Fix, in `AddActiveDomain()`:

```cpp
if (!m_loader->m_user_domains)
    m_loader->m_active_domains.clear();
```

`clear()` is immediately followed by the existing `push_back`, so the unchecked
`m_active_domains[0]` in the other branch is not newly exposed.

## Testing (Comparing RTF)

19 files, +3296/-166. RTX 4080 (CUDA 13.2, ROCm 7.2.4) and MI300X/gfx942. Configured with no GPU
flags, SCM resolves to HIP while DEM and FSI::SPH stay CUDA: 0 failures, 19 libraries, 230 demos. Note: CPU on both computers is different.

`demo_ROBOT_Viper_SCM`, 2 s headless, steady-state ms/step with warm-up excluded, both backends
confirmed to have run every step (`raycast=4001/4001 forces=4001/4001`).

RTX 4080:

| Grid | GPU | CPU | speedup | GPU, 6 obstacles | CPU, 6 obstacles | speedup |
|---|---|---|---|---|---|---|
| 10 cm | 0.33 | 0.54 | 1.6x | 0.87 | 1.56 | 1.8x |
| 5 cm | 0.60 | 1.52 | 2.5x | 1.78 | 4.30 | 2.4x |
| 2 cm | 1.50 | 7.00 | 4.7x | 6.15 | 21.6 | 3.5x |
| 1 cm | 5.25 | 26.0 | 5.0x | 22.5 | 80.6 | 3.6x |

MI300X, whose host is 1.7-2x slower, so compare speedups rather than absolute times:

| Grid | GPU | CPU | speedup | GPU, 6 obstacles | CPU, 6 obstacles | speedup |
|---|---|---|---|---|---|---|
| 10 cm | 0.64 | 0.89 | 1.4x | 1.51 | 2.92 | 1.9x |
| 5 cm | 0.93 | 2.59 | 2.8x | 2.99 | 7.23 | 2.4x |
| 2 cm | 2.21 | 13.9 | 6.3x | 11.0 | 41.3 | 3.7x |
| 1 cm | 7.99 | 50.6 | 6.3x | 38.8 | 156.8 | 4.0x |

Speedup grows with resolution because ray count scales as 1/delta^2 while the serial remainder
does not. Obstacles reduce it at fine grids: each carries an active domain covering far more
soil than the wheels do, and most of the work it adds stays on the CPU either way (contact patches,
bulldozing, erosion). At 10 cm the effect reverses, there being too few rays to saturate the GPU,
so the added parallel work is nearly free while the CPU pays for it serially. Cross-vendor agreement is far tighter than CPU-vs-GPU: GPU sinkage on the two machines
matches to 0.07% at 10 cm and 0.008% at 2 cm, CPU rows to within 0.3%.

### Open question

The GPU ray cast does not reproduce Bullet's result: 5.4-5.6 cm of wheel sinkage against 4.1-4.3 cm
through `ChCollisionSystem::RayHit`, on both vendors. It is the formulation, not the GPU, since
the CPU reference implementation of the same algorithm lands within 0.09% of the kernels. Ruled
out by measurement: the contact-force backend (~2%), the collision margin (`GetEnvelope()` is 0
for these models, and zeroing the remaining 5 mm mesh thickness moves sinkage 4.7 mm, not 12) and
precision (3 um). Which formulation is closer to reality is unresolved. The 5 cm grid is a further
loose end, giving 4.5 cm identically on both machines where the other three resolutions give 5.4
to 5.6.

### Not covered

- `demo_ROBOT_RoboSimian_SCM` checked only for correct CPU fallback, not for GPU execution.
- Run-time visualization on AMD (VSG unavailable on that cluster).
- AMD parts other than gfx942, and Windows / MSVC.


rserban:
@Auc7us  and @amd-pratmish

Keshav and Pratmish - this PR supposedly includes Pratmish's PR #755 but all commits from that PR are lost.
I am happy to look into merging this PR (and close 755), if Pratmish is ok with this "loss of credit".  
Have the two of you discussed this?

Also, there's a 2nd PR #770 - I assume that is still valid and should be considered for merge?

Please let me know.

Auc7us:
Hi @rserban and @amd-pratmish,
While Pratik's commits were lost during my development, I did credit Pratik as a co-author for the commit [0e418d8](https://github.com/projectchrono/chrono/pull/808/commits/0e418d876c999d3b75570c916d7e7ac2ca2c81ef) that includes all his changes.

rserban:
@amd-pratmish  Pratik, are you fine with merging this PR?

amd-pratmish:
@rserban: This looks good.

PR #814, opened 2026-08-27
vehicle: add held-key (game-style) keyboard driving to ChInteractiveDriver

ksha23:
Keyboard driving in `ChInteractiveDriver` is cumulative: each recorded W/A/S/D keypress nudges the corresponding input by a fixed increment, and it stays there until nudged again. Releasing a key does nothing. Measured on `demo_SEN_Gator`, holding W for 1.5 s moves the throttle target to 0.020 and holding longer does nothing more; four keypresses leave it pinned at 0.080 for the rest of the run. That ratchets, it doesn't drive, and it trips up anyone coming from a driving game where letting go of the key lets the car coast.

This adds `KeyboardMode::HELD`, in which driver inputs are a pure function of the keys currently held down: W drives the throttle target to 1 while held and to 0 on release, S drives the brake and takes precedence over the accelerator, A/D steer left/right, and steering self-centers when neither (or both) is held. The ramp toward these targets is still the same first-order dynamics `ChInteractiveDriver` already integrates (`SetGains`), so it's a function of elapsed time, not frame rate.

`CUMULATIVE` remains the default everywhere -- on `ChInteractiveDriver` itself and in what the Irrlicht and VSG vehicle visualization systems propose to an attached driver (`SetDefaultKeyboardMode`). Nothing changes out of the box for any existing demo. `HELD` is available on request: call `SetKeyboardMode()` on the driver or on either visualization system, and that explicit choice always wins over the proposed default.

Both backends absorb keyboard auto-repeat and duplicate release events through idempotent state updates, and release all keys when the render window loses keyboard focus (a release delivered to a different window would otherwise never be seen and the input would stay pinned). VSG only reports focus-out events on the XCB and Win32 backends; as of VSG 1.1.11 it never fires on macOS, so that safety net doesn't apply there (noted at the call site). VSG's `ChEventHandlerVSG` didn't forward focus events at all before this, so a small prerequisite commit adds that (empty-by-default virtuals, no existing handler changes behavior).

## Testing

Verified manually against `demo_SEN_Gator` (Irrlicht) and a Sedan demo (VSG). Irrlicht: a 1.5 s hold of W produced one press at t=0 and one release at t=1500.3 ms, tracked correctly. VSG: nine holds of W/A/S/D produced exactly nine press and nine release events, with recovered hold durations of 1499.9 ms (1500 ms hold) and 59.9 ms (60 ms tap). On a Sedan with default gains, holding W ramps the throttle to 0.3302 / 0.6329 / 0.8652 / 0.9290 / 0.9976 at 0.10 / 0.25 / 0.50 / 0.66 / 1.50 s after the press, matching `1 - exp(-4t)` to within 7e-4; on release it decays as `exp(-4t)` to 0.3677 / 0.1350 / 0.0182 at 0.25 / 0.50 / 1.00 s and reaches exactly zero. A 4x larger step size reproduces the same curve to four decimal places, confirming the ramp depends on elapsed time rather than step count.

No unit tests: `src/tests/unit_tests/vehicle` has no coverage of `ChInteractiveDriver` today (`utest_VEH_destructors.cpp` and `utest_VEH_tire_contact.cpp` don't touch it), so I didn't want to hold this change to a higher bar than what's already there. Happy to add tests for both modes if that's wanted.

### DIFF
--- a/src/chrono_vehicle/driver/ChInteractiveDriver.cpp
+++ b/src/chrono_vehicle/driver/ChInteractiveDriver.cpp
@@ -36,6 +36,13 @@ namespace vehicle {
 ChInteractiveDriver::ChInteractiveDriver(ChVehicle& vehicle)
     : ChDriver(vehicle),
       m_mode(InputMode::KEYBOARD),
+      m_keyboard_mode(KeyboardMode::CUMULATIVE),
+      m_keyboard_mode_user_set(false),
+      m_key_throttle(false),
+      m_key_brake(false),
+      m_key_steer_left(false),
+      m_key_steer_right(false),
+      m_key_clutch(false),
       m_steering_target(0),
       m_throttle_target(0),
       m_braking_target(0),
@@ -55,6 +62,7 @@ ChInteractiveDriver::ChInteractiveDriver(ChVehicle& vehicle)
 void ChInteractiveDriver::SetInputMode(InputMode mode) {
     switch (mode) {
         case InputMode::KEYBOARD:
+            ReleaseAllKeys();
             m_throttle_target = 0;
             m_steering_target = 0;
             m_braking_target = 0;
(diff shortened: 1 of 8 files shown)


rserban:
Kyle - this addition is fine and I am OK merging it.

However, the choice of behavior for keyboard controls was *deliberate*. Indeed, we are not in the business of developing a game engine, but rather simulation and analysis. As such, we are much more interested in seeing how a vehicle behaves at given, constant driver inputs than racing down a track. For instance, in your new `HELD` mode, how could you inspect vehicle behavior in a constant speed turn?

The *default* behavior should be the current one (`CUMMULATIVE` in your PR).
- please update the PR to reflect this
- please also make sure you apply clang-format to all files you touch before committing

Thanks!

ksha23:
Updated.
- CUMULATIVE is now the default everywhere, including what Irrlicht/VSG propose to an attached ChInteractiveDriver. HELD is purely opt-in via SetKeyboardMode() on the driver or either visualization system, so nothing changes out of the box for any existing demo.
- Ran clang-format over every touched file.

PR #828, opened 2026-09-05
FSI: refuse to step a fluid system that was never initialized

DanNegrut:
`ChFsiSystem::DoStepDynamics` already throws if `Initialize()` was never called;
`ChFsiFluidSystem::DoStepDynamics` does not. A standalone fluid system stepped before
`Initialize()` skips `CheckSPHParameters()` and dereferences the null `m_fluid_dynamics`. This adds
the same guard to the fluid system (same message and exception type as the wrapper), marks the
system initialized in the standalone `ChFsiFluidSystem::Initialize()` after the derived call
returns (the wrapper path already did this in `ChFsiSystem::Initialize`), and adds a unit test,
`utest_FSI-SPH_init_guard`, that checks the throw, that the rejected call changes no state, that a
system which threw can still be initialized and stepped, and that a fresh system steps.

No API change. Built alone on top of `main` and tested on CUDA (DGX Spark, CUDA 13) and HIP
(MI350X, ROCm 7.2): the new test and the six existing FSI-SPH unit tests pass on both.

## In depth description, if of interest

The wrapper's guard protects users who go through `ChFsiSystem`. The fluid system is also a public
class and is driven directly in several demos and tests, so the same misuse there reaches the null
dereference with no message. The guard is placed before the timer starts so a rejected call costs
nothing and leaves no partial state, which the test checks by stepping a fresh system, catching
the throw, then initializing and stepping the same object.

The test qualifies the call as `sysSPH.ChFsiFluidSystem::Initialize()` deliberately:
`ChFsiFluidSystemSPH` declares `Initialize` overloads that hide the base class's no-argument
`Initialize()`, so an unqualified call does not compile for a standalone user. That is a separate
matter, noted in the test's comment and not changed here.

This work used computing resources made available through the AMD University Program (AUP) AI &
HPC Cluster.

### DIFF
--- a/src/chrono_fsi/ChFsiFluidSystem.cpp
+++ b/src/chrono_fsi/ChFsiFluidSystem.cpp
@@ -17,6 +17,7 @@
 // =============================================================================
 
 #include <cmath>
+#include <stdexcept>
 #include <filesystem>
 
 #include "chrono/core/ChTypes.h"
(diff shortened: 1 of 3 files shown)


rserban:
Fixed in commit #c9fa3d14fe, using a slightly different setup (better separation of responsibilities).
Unit test not included (while useful to check correctness of the fix, such a test is not needed as a unit test).

PR #832, opened 2026-09-05
ChWheelTestRig: enlarge the estimated CRM active domain from 1.25 to 2.5 times the wheel's half-dimensions

### LINKED ISSUE #824: ChWheelTestRig CRM: the estimated active domain is too small, and the shipped wheel demo reads 12 to 43 percent low because of it
`ChWheelTestRig::SetWheelActiveDomain()` (`ChWheelTestRig.cpp:578-582` at `8601f5a89`) estimates
the box as `±1.25 * (R, W/2, R)` around the hub, and `demo_ROBOT_Viper_WheelTestRig_CRM.cpp:111`
calls it. That box reaches 1.25 wheel radii from the hub, inside the soil the wheel is shearing.
Varying only the box, at 40 percent slip, 10 mm spacing, the drawbar pull converges to about
**252 N** at twice the estimate and stays there at four and eight times. With the estimate it reads
**220 N on `e75b5024`** (1 Aug) and **147 N on current `main`**, 12 and 42 percent low; doubled,
current `main` gives 251 and 266 N. On current
`main` the estimate is also unstable: inside the one-second measurement window the force swings
between 51 and 263 N (standard deviation 46 to 54 N, against 13 to 21 N for every larger box).

Fix, one line: `1.25` to `2.5` in `SetWheelActiveDomain()`. And one sentence in its doc comment:
"double the box and rerun; if the answer moves, the box was too small."

## In depth description, if of interest

Full sweep, three repeats at scale 4, one elsewhere; both trees agree to 1.8 percent above the
estimate, so this is one default, not a physics disagreement between versions:

| box, multiple of the estimate | `e75b5024` (1 Aug) | `d107ee4` (15 Aug) and `main` |
|---|---|---|
| 1 | 220.4 N | 145.5 N (main: 148.9, 146.8) |
| 2 | 255.8 N | 251.9 N (main: 251.4, 266.4) |
| 4 | 250.6 ± 4.7 N | 255.0 ± 2.3 N |
| 8 | 250.4 N | 246.1 N |

Why the loss tripled at `d107ee4`: bisected over the sixteen commits between the two trees; the
step is exactly that commit. What changed in it that matters is not the box size, which is identical
before and after, and not the amount of soil integrated, which went up (the run costs 18 percent
more, not less). The instability suggests soil near the box boundary flipping between active and
frozen as the wheel turns. We have not established the mechanism and are not claiming one; the
default is the actionable part.

Operating-point dependence, stated so it is not over-read: at 60 percent slip the same doubling
changes the force by only 1.3 percent, but doubles the measured sensitivity to `mu_fric_2` (from
+6 to +12 percent). So the box mainly hides a parameter's effect at the shipped spacing, and mainly
hides force at lower slip.

Cost: a converged box runs about 2.8 times longer than the estimate (640 s against 227 s for a
4-second, 205,821-particle case on one GPU), which is why a too-small default is easy to keep.

Environment: DGX Spark (GB10), CUDA 13.0, fp32, `-DCHRONO_CUDA_ARCHITECTURES=121`; single-precision
run-to-run scatter 1 to 2 percent. The HIP backend on an MI350X agrees with CUDA to 2.4 percent on
identical source at this operating point.

This work used computing resources made available through the AMD University Program (AUP) AI &
HPC Cluster.

DanNegrut:
Fixes #824. `SetWheelActiveDomain()` with no argument estimated the active box as 1.25 times the
wheel's half-dimensions around the hub, which stops inside the soil the wheel is shearing; the shipped
Viper wheel demo reads 12 to 43 percent low with it and fluctuates strongly. At 2.5 times the
half-dimensions (twice the previous estimate) the drawbar pull matches the value from boxes 4 and 8
times the previous estimate (251.2 and 252.6 N against about 252 N, at 40 percent slip, 10 mm spacing,
single precision; that one operating point is where the factor was fitted, and the comments say so). One number in `ChWheelTestRig.cpp`, plus a comment there and in the
header saying how to check any box: double it and rerun.

Built on CUDA (DGX Spark, CUDA 13) and HIP (MI350X, ROCm 7.2). With the new estimate the wheel case
that read 147 N reads 256.5 and 257.6 N (on top of the other fixes) and 255.9 and 266.6 N (this change
alone on `main`), all inside the converged band.

## In depth description, if of interest

The measurements behind the number are in #824. The cost of the larger box is real (about 2.8 times
the run time of the old estimate at 205,821 particles), which is why the small default was easy to
keep; a user who wants the old size back can pass it explicitly through the `ChAABB` overload.

This work used computing resources made available through the AMD University Program (AUP) AI &
HPC Cluster.

### DIFF
--- a/src/chrono_vehicle/wheeled_vehicle/test_rig/ChWheelTestRig.cpp
+++ b/src/chrono_vehicle/wheeled_vehicle/test_rig/ChWheelTestRig.cpp
@@ -576,9 +576,16 @@ void ChWheelTestRig::SetWheelActiveDomain(const ChAABB& aabb) {
 }
 
 void ChWheelTestRig::SetWheelActiveDomain() {
+    // The box must extend well past the soil the wheel is shearing. Measured on one Viper wheel case
+    // (40 percent slip, 10 mm spacing, single precision): at 1.25 times the wheel's half-dimensions
+    // (the previous estimate) the drawbar pull read 12 to 43 percent below the value obtained with
+    // larger boxes and fluctuated strongly inside the measurement window; at 2.5 times (twice the
+    // previous estimate) it matches the values obtained with 4 and 8 times the previous estimate
+    // (5 and 10 times the wheel's half-dimensions). The factor was fitted at that one operating point.
+    // To check any box, including this one, double it and rerun: if the answer moves, it was too small.
     auto corner = ChVector3d(m_wheel_assembly->GetRadius(), m_wheel_assembly->GetWidth() / 2, m_wheel_assembly->GetRadius());
-    m_wheel_AABB.min = -1.25 * corner;
-    m_wheel_AABB.max = +1.25 * corner;
+    m_wheel_AABB.min = -2.5 * corner;
+    m_wheel_AABB.max = +2.5 * corner;
 }
 
 void ChWheelTestRig::CreateTerrainCRM() {
--- a/src/chrono_vehicle/wheeled_vehicle/test_rig/ChWheelTestRig.h
+++ b/src/chrono_vehicle/wheeled_vehicle/test_rig/ChWheelTestRig.h
@@ -262,7 +262,11 @@ class CH_VEHICLE_API ChWheelTestRig {
     void SetWheelActiveDomain(const ChAABB& aabb);
 
     /// Set a single active domain of estimated dimensions associated with the entire wheel assembly (CRM terrain only).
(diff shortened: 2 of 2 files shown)


rserban:
Issue #824 will  be addressed later, after an on-going ChWheelTestRig redesign is completed.

As a side note: documentation of a function in a public library class should not referenced a particular model, demo, or test.

